PR #2, opened 2020-02-19
Add initial contribution guidelines

jessicayuen:
Most of these are lightly modified versions of the same files from
[envoy](github.com/envoyproxy/envoy).

Add:
* Code of Conduct
* DCO
* Owners
* Cursory Style Guide
* High level contributing document

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,121 @@
+We welcome contributions from the community. Please read the following guidelines carefully to
+maximize the chances of your PR being merged.
+
+# Communication
+
+* Before starting work on a major feature, please reach out to us via GitHub, Slack,
+  email, etc. We will make sure no one else is already working on it and ask you to open a GitHub
+  issue.
+* A "major feature" is defined as any change that is > 100 LOC altered (not including tests), or
+  changes any user-facing behavior. We will use the GitHub issue to discuss the feature and come to
+  agreement. This is to prevent your time being wasted, as well as ours. The GitHub review process
+  for major features is also important so that [organizations with commit access](OWNERS.md) can
+  come to agreement on design. If it is appropriate to write a design document, the document must be
+  hosted either in the GitHub tracking issue, or linked to from the issue and hosted in a
+  world-readable location.
+* Small patches and bug fixes don't need prior communication.
+
+# Coding style
+
+* See [STYLE.md](STYLE.md)
+
+# Submitting a PR
+
+* Fork the repo.
+* Create your PR.
+* Tests will automatically run for you.
+* We will **not** merge any PR that is not passing tests.
+* PRs are expected to have 100% test coverage for added code. This can be verified with a coverage
+  build. If your PR cannot have 100% coverage for some reason please clearly explain why when you
+  open it.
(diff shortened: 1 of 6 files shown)


jessicayuen:
As this project progresses, we should also add:
* pre-commit hooks for DCO and other linting.
* rules around API versioning and deprecation.

CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
+We welcome contributions from the community. Please read the following guidelines carefully to
+maximize the chances of your PR being merged.
+
+# Communication
+
+* Before starting work on a major feature, please reach out to us via GitHub, Slack,
+  email, etc. We will make sure no one else is already working on it and ask you to open a GitHub
+  issue.
+* A "major feature" is defined as any change that is > 100 LOC altered (not including tests), or
```
lita: I would be explicit about what LOC means.

STYLE.md
```diff
@@ -0,0 +1,12 @@
+# Go coding style
+
+* The xDS Relay source code is formatted using
+  [golangci-lint](https://github.com/golangci/golangci-lint). Thus all white spaces, etc.
```
lita: Can we add all the linters we will be using here also or link to the golangci configuration file?

STYLE.md
```diff
@@ -0,0 +1,12 @@
...
+  issues are taken care of automatically. Azure pipeline tests will automatically check the code
+  format and fail.
+* More detailed descriptions of each linter can be found
+  [here](https://github.com/golangci/golangci-lint/tree/v1.23.6#enabled-by-default-linters).
```
lita: Oops, see this here. We should make our own configuration file and be explicit about what linters we want to run.

STYLE.md
```diff
@@ -0,0 +1,12 @@
...
+
+# Recommended Go style guides
+
+* [Effective Go](https://golang.org/doc/effective_go.html)
```
lita: Not saying we need to figure this out in this PR, but we should add more here. Here are some style guides to get you inspired.

Here are the Kubernetes coding conventions: https://github.com/kubernetes/community/blob/master/contributors/guide/coding-conventions.md

Uber's Go Style Guide: https://github.com/uber-go/guide/blob/master/style.md

Google's Go Code Review Comments: https://github.com/golang/go/wiki/CodeReviewComments

Feel free to pick and choose or just adopt what you think is best.

jessicayuen: I'll add Go Code Review Comments. I think with Go in particular, linters do a pretty good job of catching style smells. I'm wary of adding too many custom style guides in case of contradiction with the others. We can add more as we iterate on this project.

STYLE.md
jessicayuen: Yeah we'll definitely want a custom `.golangci.yml` file. This will be added alongside the Go scaffolding and I can update this style guide to point to it once its in place. I think it makes more sense to be done in a future PR, lmk.

STYLE.md
lita: Ah yeah. I wasn't asking to put all of these in here. I think we should also have a section of our own and take what we want! :)

PR #11, opened 2020-02-26
Bootstrap Go project and add logging framework

jessicayuen:
Bootstrap the initial project via the following, 

* Add `.gitignore`
* Add Makefile
* Vendor dependencies using go modules
* Add logging framework
* Add preliminary Go project dir structure and cursory go-control-plane server scaffolding


jessicayuen:
I'll make a follow-up PR to configure the Docker image and Azure Pipelines to run builds and test.

internal/app/server/server.go
```diff
@@ -0,0 +1,54 @@
...
+
+package server
+
+import (
+	"context"
+	"net"
+
+	"github.com/envoyproxy/xds-relay/internal/pkg/log"
+
+	api "github.com/envoyproxy/go-control-plane/envoy/api/v2"
```
jyotimahapatra: use `v3` instead?

internal/app/server/server.go
```diff
@@ -0,0 +1,54 @@
...
+	api.RegisterClusterDiscoveryServiceServer(server, gcpServer)
+	api.RegisterRouteDiscoveryServiceServer(server, gcpServer)
+	api.RegisterListenerDiscoveryServiceServer(server, gcpServer)
+
+	logger.Info("Initializing server at", listener.Addr().String())
+	go func() {
+		if err := server.Serve(listener); err != nil {
+			logger.Fatalw("failed to initialize server", "err", err)
+		}
+	}()
```
jyotimahapatra: does the main loop exit and so does the process. We have to add some kind of `waitGroup.wait()` logic here.

jessicayuen: Yes, let me take the server out of the Go routine and serve this on the main thread.

jyotimahapatra: or may be leave the server out in the initial PR to unblock other streams? We do need to add some other aspects to serve also like graceful shutdown as well at a later point.

internal/app/server/server.go
jessicayuen: Chatted offline. go-control-plane isn't in a state where it supports v3 APIs. @jyotimahapatra will open an issue and link it here.

internal/app/server/server.go
jessicayuen: It shouldn't block other streams, the cache / mapper / control plane client are primarily interfaces. We can always serve additional non-blocking workflows on separate Go routines. I think the server should run on the main thread, but let me know if I missed something.

jyotimahapatra: ah..i meant the workstreams to parallelize the effort once the scaffolding is in.

go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/envoyproxy/xds-relay
+
+go 1.12
+
+require (
+	github.com/envoyproxy/go-control-plane v0.9.4
+	github.com/onsi/ginkgo v1.12.0
+	github.com/onsi/gomega v1.9.0
+	go.uber.org/zap v1.14.0
+	google.golang.org/grpc v1.27.1
```
lita: Do we need to version the grpc version? does go-control-plane do that for us?

lita:
What is the plan for making sure that go-control-plane stays up to date here?

Makefile
```diff
@@ -0,0 +1,18 @@
+export SERVICE_NAME=xds-relay
+
+# Compiles the binary and installs it into /usr/local/bin
+.PHONY: compile
+compile:
+	mkdir -p ./bin && \
+	  go build -o ./bin/${SERVICE_NAME} && \
+	  cp ./bin/${SERVICE_NAME} /usr/local/bin/${SERVICE_NAME}
```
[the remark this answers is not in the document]
lita: or convert everything to 2 spaces. it would be good to be consistent

internal/app/server/server.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// Run instantiates a running gRPC server for accepting incoming xDS-based
+// requests.
+func Run() {
+	logger := log.New().Sugar()
+
+	// Cursory implementation of go-control-plane's server.
+	snapshotCache := cache.NewSnapshotCache(false, cache.IDHash{}, nil)
+	gcpServer := gcp.NewServer(context.Background(), snapshotCache, nil)
```
lita: Shouldn't we pass in a WithCancel context so that we can shutdown via SIGINT?

go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/envoyproxy/xds-relay
+
+go 1.12
```
eapolinario: why so old? Does this have to go in lockstep with `go-control-plane`?

Makefile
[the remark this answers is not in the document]
jessicayuen: Tabs are purposeful unfortunately,

> The tab character is a very important part of the syntax of Makefiles. All command lines must start with tabs.

source: https://stackoverflow.com/questions/2131213/can-you-make-valid-makefiles-without-tab-characters

With regards to the remainder of whitespacing, there are some suggested guidelines here:
https://openjdk.java.net/groups/build/doc/code-conventions.html

but mainly:
> Do not use tabs, only spaces (except for in recipes, as stated by rule 3-7).

FWIW this is new to me too, but also a practice I see followed in Lyft internal Makefiles. :)

go.mod
jessicayuen: No there shouldn't be any reason it can't be more recent. This is just what it defaulted to when I ran `go mod init`. Could be my version of Go, I'll upgrade this to 1.14.

go.mod
jessicayuen: We need to instantiate our own gRPC server (which gets passed to go-control-plane), so we do need to track our own versioning:
https://github.com/envoyproxy/xds-relay/pull/11/files#diff-6bf74a2dac7bcac1a44aa2e114bada9cR37

internal/app/server/server.go
jessicayuen: Let me update the PR to include `WithCancel`.

This is a very cursory implementation of the server. Only meant to place it to unblock some parallel work streams. i.e., We won't want to be using the snapshot cache in the final implementation either.

lita:
LGTM +1

internal/app/server/server.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2020 Envoyproxy Authors
```
mattklein123: We don't need to have license text at the top of every file, so I would advise deleting it and not adding it to new files, just an FYI.

jessicayuen: Just saw this, was trying to be consistent with go-control-plane, but I can remove them :)

PR #14, opened 2020-03-02
Add protos definition for aggregation rules

eapolinario:
This PR adds the first definition of the aggregation rules defined in the xds-relay [spec](https://docs.google.com/document/d/1X9fFzqBZzrSmx2d0NkmjDW2tc8ysbLQS9LaRQRxdJZU/edit#heading=h.dx1um2qdkbli).

We use [protoc-gen-validate](https://github.com/envoyproxy/protoc-gen-validate) to enforce rules on the proto definitions.

We follow Google's [guidance](https://cloud.google.com/apis/design/versioning#guidance) to version the proto definitions. A very similar model is also adopted by [envoy](https://github.com/envoyproxy/envoy/blob/master/api/API_VERSIONING.md#api-semantic-versioning). 

Please note that we're also following go's unofficial convention of putting the compiled proto files under the `pkg` directory, following the same naming scheme used to version the proto definitions.


api/aggregation.proto
```diff
@@ -0,0 +1,181 @@
...
+    RequestNodeFragment request_node_fragment = 2;
+
+    // A fragment generated from Envoy resource names.
+    ResourceNamesFragment resource_names_fragment = 3;
+
+    // A simple string fragment
+    string string_fragment = 4;
+  }
+
+  message RepeatedResultPredicate {
```
eapolinario: This message is the only real difference between the spec and what the proto compiler accepts. I was seeing this error in:
```
aggregation.proto:164:9: Fields in oneofs must not have labels (required / optional / repeated).
```

api/aggregation.proto
```diff
@@ -0,0 +1,181 @@
...
+    oneof type {
+      option (validate.required) = true;
+
+      // Match on the extract string value.
+      string exact_match = 2;
+
+      // Match on a regex pattern.
+      // TODO potentially use "safe regex"
+      // https://github.com/envoyproxy/envoy/blob/10f756efa17e56c8d4d1033be7b4286410db4e01/api/envoy/type/matcher/v3/regex.proto
+      string regex_match = 3;
```
lita: Why can't w use regex now?

eapolinario: No reason, I was just copying the proto definition from @jessicayuen 's spec, but I think we can commit to use envoy's regex matcher from the get go.

jessicayuen: Could you investigate into what that would look like @eapolinario? It should be relatively simple, the main thing is to ensure that the constraints set by safe regex don't limit what we may want to match on for a request node field (ex: wildcard prefixes).

jessicayuen:
Can you label this `[WIP]` so we can ignore the DCO and things in the meanwhile?

We probably want to move the proto under `/api/v1/` @mattklein123 might have some thoughts on how to structure API versioning.

jessicayuen:
> how to hook up CI?

I suggest adding a `api.yml` to the `.github.com/workflows` directory to run CI on any files in the `api/` path. Lyft internally has a pretty good example of using protoc-gen-validate in adhocinfra that I would refer to when setting this up.


>  where should the generated protos live? I put them under generated/ but there might be some convention I'm missing.

How about the following?

- `/api/proto/v1/*.proto` for `.proto` files. (We'll probably use swagger for API definitions so this layout gives flexibility for that)

- `/pkg/api/v1/` for `.pb.go` files. (`pkg` in Go has conventionally been used to express library availability for external users to consume. The generated files makes sense here if we were to build external tooling around validation of the aggregation key).

> A follow up to the previous question: do we have to commit the generated .go files?

Yes, since they'll be referenced in our source code.

Makefile
```diff
@@ -9,10 +9,32 @@ compile:
...
 # Run all unit tests with coverage report
 .PHONY: unit
 unit:
 	go test -v -cover ./...
+
+.PHONY: install-protoc
+install-protoc:
+# Rely on homebrew in the case of MacOSX for now since CI is going to be running only Linux
+ifeq ($(shell uname -s), Darwin)
+	brew install protobuf
```
jyotimahapatra: can we provide an error message if brew is not installed?

Makefile
```diff
@@ -9,10 +9,32 @@ compile:
...
+else
+	apt-get update && apt-get install unzip
+	wget -O /tmp/protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.6.1/protoc-3.6.1-linux-x86_64.zip  \
```
jyotimahapatra: just a nit: starts with`&&` here for consistency with other lines.

Makefile
eapolinario: It'll be pretty clear that `brew` is not installed, i.e., in case `brew` is not installed you're going to see something like `make: breww: No such file or directory`

scripts/generate-api-protos.sh
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/env bash
```
jyotimahapatra: May be this aspect is already handled: The developers are supposed to make changes to the proto files and commit the generated files in the PR. What happens if they don't. Can we put some check such that there are no git diffs in the generated proto files?

go.mod
```diff
@@ -1,11 +1,18 @@
...
 	github.com/envoyproxy/go-control-plane v0.9.4
-	github.com/onsi/ginkgo v1.12.0
-	github.com/onsi/gomega v1.9.0
-	go.uber.org/zap v1.14.0
-	google.golang.org/grpc v1.27.1
+	github.com/envoyproxy/protoc-gen-validate v0.3.0
+	github.com/go-swagger/go-swagger v0.22.0 // indirect
+	github.com/golang/protobuf v1.3.4
+	github.com/grpc-ecosystem/grpc-gateway v1.13.0 // indirect
+	github.com/iancoleman/strcase v0.0.0-20191112232945-16388991a334 // indirect
```
jessicayuen: Can we remove the indirect packages?

go.mod
```diff
@@ -1,11 +1,18 @@
...
+	github.com/lyft/protoc-gen-star v0.4.14 // indirect
+	github.com/onsi/ginkgo v1.7.0
+	github.com/onsi/gomega v1.4.3
+	go.uber.org/zap v1.13.0
+	google.golang.org/grpc v1.25.1
```
jessicayuen: Revert the version bump downs.

jessicayuen:
@eapolinario Please tag me for review when you're ready for me to take a more comprehensive look!

scripts/generate-api-protos.sh
```diff
@@ -0,0 +1,195 @@
...
+    "https://github.com/angular/clang-format/raw/v1.4.0/bin/${clang_format_os}_x64/clang-format"
+  md5check "${clang_format_md5}" "/tmp/clang-format"
+  chmod a+x "/tmp/clang-format"
+  mv "/tmp/clang-format" "${CLANG_FORMAT_BIN}"
+
+fi
+
+# Install or update protobuf.
+if [[ -f /etc/alpine-release ]]; then
+  # Alpine must use it's built-in version of protoc since the published version uses glibc and does not support musl.
```
jyotimahapatra: do we use alpine docker image to run the script?

scripts/generate-api-protos.sh
eapolinario: Good call, I added a script to check for this.

go.mod
eapolinario: actually we can't, these are pulled in by the calls to `go get` in the `generate-api-protos` script.

.github/workflows/api.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: ${{ env.PACKAGEPATH }}
+      - uses: actions/setup-go@v2-beta
+        with:
+          go-version: '^1.13.1'
+      - name: Compile protobufs
+        working-directory: ${{ env.PACKAGEPATH }}
+        run: ./scripts/generate-api-protos.sh
```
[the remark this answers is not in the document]
eapolinario: good call.

api/aggregation.proto
eapolinario: We talked offline about this and we reached the conclusion that instead of relying on envoy's regex matcher type we should define our own, however, this PR is already gigantic as is, so... how about we tackle this in follow up PR?

scripts/generate-api-protos.sh
eapolinario: I removed all the alpine-specific code.

jyotimahapatra:
@eapolinario I got the branch and ran `make compile`
```
➜  xds-relay git:(proto) make compile-protos
./scripts/generate-api-protos.sh
info: Downloading protoc-v3.11.4 to build environment
go list -m: not using modules
make: *** [compile-protos] Error 1
➜  xds-relay git:(proto) pwd
/Users/jmahapatra/go/src/github.com/envoyproxy/xds-relay
```
do i have to run any setup commands?

api/xdsrelay/aggregation/v1/aggregation.proto
```diff
@@ -0,0 +1,182 @@
+syntax = "proto3";
```
jessicayuen: Could we keep the package in `/api/protos/aggregation/v1/aggregation.proto`?

pkg/xdsrelay/aggregation/v1/aggregation.pb.go
```diff
@@ -0,0 +1,994 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
jessicayuen: Could we move this to `/pkg/api/aggregation/v1`?

.github/workflows/api.yml
```diff
@@ -0,0 +1,34 @@
...
+      - 'api/**'
+      - .github/workflows/api.yml
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'api/**'
+      - .github/workflows/api.yml
+env:
+  GOPATH: ${{ github.workspace }}/go
```
jessicayuen: I think you can delete this envvar since it's unused

scripts/api-ensure-no-diff.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
jessicayuen: Could you put a high level comment for what this script does?

What do you think of naming this script: `ensure-protos-compiled.sh`?

go.mod
jessicayuen: User has to manually commit the changes in this file though right? I would just go in and delete the indirects.

jessicayuen: cc @twoism, @rodaine here for advice on best Go practices

scripts/generate-api-protos.sh
```diff
@@ -0,0 +1,116 @@
...
+    --plugin=protoc-gen-validate="${GOBIN}/protoc-gen-validate" \
+    --validate_out="lang=go:pkg/" \
+    --plugin=protoc-gen-go="${GOBIN}/protoc-gen-go" \
+    "${1}"
+}
+
+main() {
+  # Install required plugins for generation. These deps are pinned in go.mod.
+  go install \
+    github.com/envoyproxy/protoc-gen-validate \
```
jyotimahapatra: can we also put a go version check to ensure version 1.14? I was trying this out and figured `make compile-protos` and `make compile` were giving me messages that were hard to comprehend.

.github/workflows/api.yml
```diff
@@ -0,0 +1,37 @@
...
+env:
+  PACKAGEPATH: ${{ github.workspace }}/go/src/github.com/${{ github.repository }}
+jobs:
+  protobuf-generate:
+    strategy:
+      matrix:
+        go-version: [1.14.x]
+        platform: [ubuntu-latest, macos-latest]
+    name: protobuf compile check
+    runs-on: ubuntu-latest
```
jyotimahapatra: while using the matrix, do we still have to use the `runs-on`?

eapolinario: oops.

eapolinario: [runs-on](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on) is required. This is the error in case we remove it: https://github.com/envoyproxy/xds-relay/pull/14/checks?check_run_id=488735330

pkg/xdsrelay/aggregation/v1/aggregation.pb.go
eapolinario: done.

scripts/generate-api-protos.sh
jyotimahapatra: yeah, parsing might break between verrsions..but may be a check for 1.14 in the response

eapolinario: I feel this is too brittle. How about we handle this in the documentation? Maybe make it very explicit in the main `README`?

jyotimahapatra: sounds good.

scripts/api-ensure-no-diff.sh
eapolinario: done.

api/protos/aggregation/v1/aggregation.proto
```diff
@@ -0,0 +1,179 @@
...
+// [#next-free-field: 2]
+message KeyerConfiguration {
+
+  // [#next-free-field: 2]
+  message Fragment {
+
+    // A rule defining how to match a Envoy request and what resulting
+    // fragment to generate.
+    // [#next-free-field: 3]
+    message Rule {
```
jyotimahapatra: @eapolinario qs for my understanding: what is the criteria for choosing when to keep a proto message within another message or outside?
I see that Rule is defined inside Fragment and others are outside.

jessicayuen: Majority of this from the spec so I'll shed some light on why I went about it this way. The ones that are inside the proto message are contained, similar to Java inner classes. They won't be used by other outer protos. The ones defined outside are shared.

jyotimahapatra: Thank you for explaining

api/protos/aggregation/v1/aggregation.proto
```diff
@@ -0,0 +1,179 @@
...
+
+      // Defines how to match an Envoy Request.
+      MatchPredicate match = 1 [(validate.rules).message.required = true];
+
+      // Defines how to generate the resulting fragment if matched.
+      ResultPredicate result = 2 [(validate.rules).message.required = true];
+    }
+
+    // The first rule that matches will be applied in sequential order.
+    repeated Rule rules = 1 [(validate.rules).repeated .min_items = 1];
```
jyotimahapatra: there;s a space between `repeated` and `.min_items`.

Issue #7, opened 2020-02-25
Aggregation rules bootstrap file validator


jessicayuen:
Define a standalone process / tool for validating the `.yaml` bootstrap configuration in #6 .

This tooling is meant to be run by end users of xDS Relay in their CI or test systems.

Envoy has a [route checker tool](https://github.com/envoyproxy/envoy/tree/master/test/tools/router_check) that can be used for inspiration.

PR #17, opened 2020-03-11
Implement Mapper.GetKey

jyotimahapatra:
Implementing the Mapper.GetKey 

The responsibility of the GetKey api is to map each incoming request to an aggregated key . 
The aggregated key is used to query the backing control plane and maintain cache of xds watches.


jyotimahapatra:
@jessicayuen @eapolinario Im working on adding more complicated tests now. But you could start dumping your initial thoughts now

jessicayuen:
@jyotimahapatra I'll leave the choice up to you, but feel free to support the additional fragment types in separate PRs. This is a meaty component and I don't expect it to all be done together.

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,228 @@
...
+	"regexp"
+
+	core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
+	aggregationv1 "github.com/envoyproxy/xds-relay/pkg/api/aggregation/v1"
+)
+
+type Rule = aggregationv1.KeyerConfiguration_Fragment_Rule
+
+type Mapper interface {
+	GetKeys(node core.Node, typeURL string) (string, error)
```
jessicayuen: I think this param should just be a [v2.DiscoveryRequest](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/discovery.proto#discoveryrequest), since it'll contain the node, type url, resource information, and additional fragment types that we would like to support in the future.

jessicayuen: Singular `GetKey`, since any request will only map to the first matched aggregated key.

jyotimahapatra: i think we should not be working with DiscoveryRequest too deep into our stack since its a transport object. Its better to remove dependency and work with just the resources

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,255 @@
...
+	core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
+	aggregationv1 "github.com/envoyproxy/xds-relay/pkg/api/aggregation/v1"
+)
+
+type rule = aggregationv1.KeyerConfiguration_Fragment_Rule
+type matchPredicate = aggregationv1.MatchPredicate
+type repeatedResultPredicate = aggregationv1.ResultPredicate_RepeatedResultPredicate
+
+// Mapper defines the interface that Maps an incoming request to an aggregation key
+type Mapper interface {
```
jessicayuen: It may be worth moving the interface to the external `/pkg` directory so that others can define their own mapper implementation if they choose to.

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,255 @@
...
+	GetKey(node core.Node, typeURL string) (string, error)
+}
+
+type mapper struct {
+	config aggregationv1.KeyerConfiguration
```
jessicayuen: would use the logger defined in `/internal/pkg/log` and store it here, replacing the `fmt` calls.

(I'm separately considering placing an interface in front of the zap logger but that's a different conversation)

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,255 @@
...
+		for _, fragmentrule := range fragmentrules {
+			matchpredicate := fragmentrule.GetMatch()
+			isMatch, err := isMatchPredicate(matchpredicate, node, typeURL)
+			if err != nil {
+				return "", err
+			}
+			if isMatch {
+				return getResult(fragmentrule, node)
+			}
+		}
```
jessicayuen: I think we're missing appending of the other fragments.
i.e. if the keyer configuration had 3 fragments, we're returning
`<fragment_result_1>` rather than,
`<fragment_result_1>_<fragment_result_2>_<fragment_result_3>`

We still pick the first rule match, but need to consider the other fragments as well.

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,255 @@
...
+		return isOrMatch(matchPredicate, node, typeURL)
+	} else if matchPredicate.GetNotMatch() != nil {
+		return isNotMatch(matchPredicate, node, typeURL)
+	} else if matchPredicate.GetRequestTypeMatch() != nil {
+		return isRequestTypeMatch(matchPredicate, typeURL), nil
+	} else if matchPredicate.GetRequestNodeMatch() != nil {
+		return isNodeTypeMatch(matchPredicate, node)
+	}
+
+	return false, nil
```
jessicayuen: Thoughts on error-ing here for an unsupported predicate? Or warn logging?

internal/pkg/mapper/mapper.go
jyotimahapatra: also ADS depends on typeurl being empty. The typeurl inside DiscoveryRequest might not be enough to make the decision.

internal/pkg/mapper/mapper.go
jyotimahapatra: `fmt` is used to create `Error` objects. Will look at zap to see if it supports a way to create error objects.

internal/pkg/mapper/mapper.go
jyotimahapatra: yes...there's no logging now. Will add logs today

internal/pkg/mapper/mapper.go
jyotimahapatra: yes i considered that, but did not take the approach in order to be opinionated about rule parsing in xds-relay. Would we not improve the rules rather than exposing the interface on feature asks?

internal/pkg/mapper/mapper.go
jyotimahapatra: right. Fixing this today

internal/pkg/mapper/mapper.go
jyotimahapatra: On an unsupported predicate qs, we should also be doing `Validate()` to check consistency of the proto.

jyotimahapatra:
@jessicayuen @eapolinario Do you think the PR is large for review and you could benefit from smaller PRs to understand the flow better? LMK and i can implement one predicate each in separate PRs

jessicayuen:
> @jessicayuen @eapolinario Do you think the PR is large for review(250 lines in mapper.go) and you could benefit from smaller PRs to understand the flow better? LMK and i can implement one predicate each in separate PRs

I think what you have is fine, it was pretty easy to follow. I was referring to additional fragments,
> Todo:
[] implement ResourceNamesFragment proto

jyotimahapatra:
@jessicayuen @eapolinario Sure, im parking `ResourceNamesFragment` for another iteration. 
The test files has grown to ~800 lines and let me know if you have trouble navigating the tests and I can split the PR for the purpose of making tests easy to review.

jyotimahapatra:
I'm going to split this up into separate PRs for ease of further discussion.

jyotimahapatra:
#7

Issue #9, opened 2020-02-25
Implement State-of-the-world xDS cache API


jessicayuen:
Refer to the tech spec on [Cache](https://docs.google.com/document/d/1X9fFzqBZzrSmx2d0NkmjDW2tc8ysbLQS9LaRQRxdJZU/edit#heading=h.e7gk6wgvxz9d).

xDS Relay will maintain its own in-memory cache to keep the most recent response given the aggregated key.

Each key will maintain the following fields:
- _resp_: the most recent discovery response from the control plane
- _watchers_: The open watches (requests) awaiting a response.
- _stream_open_: Identifies whether there is already a connection open with go-control-plane for this aggregated key. If the value is false, the first request will be propagated to the control plane.

Implement Cache interface and API for the state-of-the-world xDS Requests. This is a proposed set of APIs.
```
Exists(key: string): bool
- Return false if the key exists.

Fetch(key: string): *v3.Response
- Return the cached response if it exists, otherwise nil.

IsStreamOpen(key: string): bool
- Return false if no stream is opened against the upstream control plane for the given key.

SetStreamOpen(open: bool)
- Sets whether there is a stream open against the upstream control plane.

Set(resp: v3.Response, key: string): ([]*v3.Request, err)
- Set the cache response and return the list of open watches.

Clear(key: string)
- Discards the existing watches. This is meant to be used in tangent with the API above.

Set(req: v3.Request, key: string): (bool, err)
- Adds the watch to the cache. Returns whether a stream is open.
```

LisaLudique:
Exists should return true if the key exists, right?

jessicayuen:
> Exists should return true if the key exists, right?
yes! typo.

Feel free to modify these API suggestions as you see fit. I would primarily consider the sequence diagram in https://docs.google.com/document/d/1X9fFzqBZzrSmx2d0NkmjDW2tc8ysbLQS9LaRQRxdJZU/edit#heading=h.1g62tk7rs8tn, to make sure the cache interface allows for the request manager to issue this workflow.

PR #18, opened 2020-03-12
Initial cache interface

LisaLudique:
This change defines the cache interface for an in-memory cache that retains the DiscoveryResponses from the control plane and related watches/stream for a given aggregated key. The methods were defined based on the [sequence diagram](https://docs.google.com/document/d/1X9fFzqBZzrSmx2d0NkmjDW2tc8ysbLQS9LaRQRxdJZU/edit#heading=h.1g62tk7rs8tn) from the tech spec.

Issue: #9

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 bin/
 # third-party dependencies via go mod
 vendor/
-
 # ginkgo test coverage files
 *.coverprofile
+# IntelliJ settings
+.idea/
--- a/internal/app/cache/cache.go
+++ b/internal/app/cache/cache.go
@@ -11,12 +11,6 @@ type Cache interface {
 	// Fetch returns the cached response if it exists.
 	Fetch(key string) (*envoy_api_v2.DiscoveryResponse, error)
 
-	// IsStreamOpen returns false if no stream is opened against the upstream control plane for the given key.
-	IsStreamOpen(key string) (bool, error)
-
-	// SetStreamOpen sets whether there is a stream open against the upstream control plane for the given key.
-	SetStreamOpen(key string, open bool) error
-
 	// SetResponse sets the cache response and returns the list of open watches.
 	SetResponse(key string, resp envoy_api_v2.DiscoveryResponse) ([]*envoy_api_v2.DiscoveryRequest, error)
 
--- a/internal/app/cache/cache.go
+++ b/internal/app/cache/cache.go
@@ -1,7 +1,7 @@
 package cache
 
 import (
-	envoy_service_discovery_v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	envoy_api_v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 )
 
 type Cache interface {
@@ -9,7 +9,7 @@ type Cache interface {
 	Exists(key string) bool
 
 	// Fetch returns the cached response if it exists.
-	Fetch(key string) (*envoy_service_discovery_v3.DiscoveryResponse, error)
+	Fetch(key string) (*envoy_api_v2.DiscoveryResponse, error)
 
 	// IsStreamOpen returns false if no stream is opened against the upstream control plane for the given key.
 	IsStreamOpen(key string) (bool, error)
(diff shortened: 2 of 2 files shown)


internal/app/cache/cache.go
```diff
@@ -0,0 +1,31 @@
...
+	Exists(key string) bool
+
+	// Fetch returns the cached response if it exists.
+	Fetch(key string) (*envoy_service_discovery_v3.DiscoveryResponse, error)
+
+	// IsStreamOpen returns false if no stream is opened against the upstream control plane for the given key.
+	IsStreamOpen(key string) (bool, error)
+
+	// SetStreamOpen sets whether there is a stream open against the upstream control plane for the given key.
+	SetStreamOpen(key string, open bool) error
```
LisaLudique: Because the setting of the stream seems to always be tied to other actions (setting response, adding watch), I don't think this method is needed in the interface.

internal/app/cache/cache.go
```diff
@@ -0,0 +1,31 @@
+package cache
+
+import (
+	envoy_service_discovery_v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+)
+
+type Cache interface {
```
LisaLudique: 1) Added error to the return values for several of these methods. It may not be necessary, but while I was working with freecache I found that we might need to propagate errors corresponding to being unable to decode the cache value, encode a cache value, the key not existing, etc.

2) Not sure how to add this to a go interface but the cache should be instantiated with TTL and memory as configurable values.

internal/app/cache/cache.go
```diff
@@ -0,0 +1,31 @@
+package cache
+
+import (
+	envoy_service_discovery_v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
```
jyotimahapatra: may be use v2 for now? Once we have the v2/v3 story figured in g-c-p we can introduce the same technique here.

LisaLudique: I'm fine with using v2 here, but the spec definitions all use v3 types. Have we decided that the MVP of ids-relay will use v2, unless we do the v2/v3 conversion before we finish the MVP?

jyotimahapatra: I think we should be making sure xds-relay can talk to g-c-p and that means going with v2 as of now. we are going to support both versions. for now the next work stream to connect to control plane might get blocked because of using v3.

internal/app/cache/cache.go
```diff
@@ -0,0 +1,31 @@
...
+
+type Cache interface {
+	// Exists returns true if the key exists in the cache, false otherwise.
+	Exists(key string) bool
+
+	// Fetch returns the cached response if it exists.
+	Fetch(key string) (*envoy_service_discovery_v3.DiscoveryResponse, error)
+
+	// IsStreamOpen returns false if no stream is opened against the upstream control plane for the given key.
+	IsStreamOpen(key string) (bool, error)
```
jyotimahapatra: both  `IsStreamOpen` and `SetStreamOpen` related functions are probably internal to implementation. If the cache is a write through cache the g-c-p connection will be handled by the implementation.

LisaLudique: I agree (https://github.com/envoyproxy/xds-relay/pull/18#discussion_r391852472) - I removed both from the public interface.

internal/app/cache/cache.go
```diff
@@ -0,0 +1,25 @@
...
+	Exists(key string) bool
+
+	// Fetch returns the cached response if it exists.
+	Fetch(key string) (*envoy_api_v2.DiscoveryResponse, error)
+
+	// SetResponse sets the cache response and returns the list of open watches.
+	SetResponse(key string, resp envoy_api_v2.DiscoveryResponse) ([]*envoy_api_v2.DiscoveryRequest, error)
+
+	// ClearWatches discards the existing watches.
+	ClearWatches(key string) error
```
jyotimahapatra: @LisaLudique Do we envision ClearWatches to be called in the scenario of TTL expiration?

internal/app/cache/cache.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// AddWatch adds the watch to the cache and returns whether a stream is open.
+	AddWatch(key string, req envoy_api_v2.DiscoveryRequest) (bool, error)
+
+	// IsWatchPresent returns whether the given request is watching for the resource corresponding to the given key.
+	IsWatchPresent(key string, req envoy_api_v2.DiscoveryRequest) (bool, error)
```
jessicayuen: Hmm where did you envision us using this API?

internal/app/cache/cache.go
jessicayuen: 1. Makes sense
2. I don't believe these need to be a part of the interface. They can be implementation details in the internal struct implementation when constructing the Cache object.

.gitignore
```diff
@@ -2,6 +2,7 @@
 bin/
 # third-party dependencies via go mod
 vendor/
-
 # ginkgo test coverage files
 *.coverprofile
+# IntelliJ settings
```
jessicayuen: nit: when you merge in the conflict from jyoti's change, can you just lump these under `#IDE settings`?

internal/app/cache/cache.go
LisaLudique: No, for TTL expiration I imagine we'd clear the entire entry for a key, not just the watches field. I thought this field would be relevant when we get a DiscoveryResponse from the control plane - per the sequence diagram and #9, this method is used to discard to watches when a response is set for a given key.

However, I can see this method also being built into SetResponse, so we may not need it as a separate public method.

internal/app/cache/cache.go
LisaLudique: This method was meant to correspond to "IS_WATCHING(D)" in the sequence diagram - my understanding is that if there is a cache hit and the stream is open, the watch can be appended.

If there's no problem with having duplicate watches, this method can be nixed.

PR #19, opened 2020-03-12
mapper interface

jyotimahapatra:
Creates the mapper interface based on [design](https://docs.google.com/document/d/1X9fFzqBZzrSmx2d0NkmjDW2tc8ysbLQS9LaRQRxdJZU/edit#).

The responsibility of the interface is to map an incoming xds request into an aggregated key.

PR contents:
1. interface
2. implementation with empty method
3. one sample test to kickoff ci for mapper


jyotimahapatra:
@jessicayuen @eapolinario @LisaLudique PTAL

jyotimahapatra:
#7

internal/pkg/mapper/mapper_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package mapper_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestLog(t *testing.T) {
```
jessicayuen: `TestMapper`?

jessicayuen: Feel free to not use this BDD framework btw (per Chris' suggestion).

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"fmt"
+
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	aggregationv1 "github.com/envoyproxy/xds-relay/pkg/api/aggregation/v1"
+)
+
+// Mapper defines the interface that Maps an incoming request to an aggregation key
+type Mapper interface {
```
jyotimahapatra: @jessicayuen continuing the [conversation](https://github.com/envoyproxy/xds-relay/pull/17#discussion_r391422717) here. I feel that the Mapper impl should closely match the proto and the spec defined in the doc. I don't foresee how Mapper can be overridden without diluting the spec from the design doc.

internal/pkg/mapper/mapper_suite_test.go
jyotimahapatra: The framework works great due to the `DescribeTable` feature!! It turned out extremely easy to write same test case with multiple inputs. In usual `testing` library in go, one has to do more boilerplate.

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,29 @@
...
+	GetKey(request v2.DiscoveryRequest, typeURL string) (string, error)
```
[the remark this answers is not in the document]
jyotimahapatra: I'm going back to DiscoveryRequest to emulate g-c-p. Also i'm still passing in typeUrl to handle ADS scenario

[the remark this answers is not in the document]
jyotimahapatra: Once we figure the v2/v3 stuff in g-c-p, we can try not using DiscoveryRequest, but as of now trying not to overthink

internal/pkg/mapper/mapper.go
jessicayuen: makes sense, I think if we ever wanted to make this more customizable, it would be worth another spec / deeper discussion.

jyotimahapatra: makes sense.

internal/pkg/mapper/mapper_suite_test.go
jyotimahapatra: The golang boilerplate looks like this https://github.com/google/cel-go/blob/master/interpreter/interpreter_test.go#L764-L776

jyotimahapatra:
@jessicayuen Once this gets merged i will bring in the predicates one at a time in subsequent PRs for reviewers to easily follow.

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,30 @@
...
+type Mapper interface {
+	// GetKey converts a request into an aggregated key
+	GetKey(request v2.DiscoveryRequest, typeURL string) (string, error)
+}
+
+type mapper struct {
+	config aggregationv1.KeyerConfiguration
+}
+
+// NewMapper construts a concrete implementation for the Mapper interface
```
LisaLudique: nit: constructs

internal/pkg/mapper/mapper_test.go
```diff
@@ -0,0 +1,88 @@
...
+	nodecluster    = "cluster"
+	noderegion     = "region"
+	nodezone       = "zone"
+	nodesubzone    = "subzone"
+	stringfragment = "stringfragment"
+)
+
+var _ = Describe("GetKey", func() {
+	DescribeTable("should be able to return error",
+		func(match *MatchPredicate, result *ResultPredicate, typeurl string, assert string) {
```
LisaLudique: nit: camelcase typeurl

jyotimahapatra: yeah...the go linter likes it this way.

jyotimahapatra: image

LisaLudique: typeURL is fine too, but does it enforce all lowercase in that test function?

jyotimahapatra: which function are you referring to?

LisaLudique: The one for this GitHub comment (L32)

LisaLudique: it's not a big deal

jyotimahapatra: image

jyotimahapatra: may be specific to my vscode only?

jessicayuen: yeah `typeURL` would be the go preferred option

jyotimahapatra: i think as long as the `make lint` goes through, we should be ok

jyotimahapatra: i understand its hard to catch such typos, so lets lean in to the linter and do what it tells

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	aggregationv1 "github.com/envoyproxy/xds-relay/pkg/api/aggregation/v1"
+)
+
+// Mapper defines the interface that Maps an incoming request to an aggregation key
+type Mapper interface {
+	// GetKey converts a request into an aggregated key
+	GetKey(request v2.DiscoveryRequest, typeURL string) (string, error)
```
jessicayuen: I have some concerns about including typeURL here because there is a typeURL in the discovery request. How do we handle ambiguity when it's supplied in both places? ADS will still wrap the discovery request no? https://github.com/envoyproxy/envoy/blob/master/api/envoy/service/discovery/v2/ads.proto#L27

jyotimahapatra: https://github.com/envoyproxy/go-control-plane/blob/master/pkg/server/server.go#L53
TypeUrl will be "" for ADS. and then we have these kind of sspecial casing for ADS https://github.com/envoyproxy/go-control-plane/blob/master/pkg/server/server.go#L305-L311

https://github.com/envoyproxy/envoy/blob/master/api/envoy/api/v2/discovery.proto#L43-L46

All of these suggest that typeurl could be empty. We probably do not want the concept of typeUrl to leak into other components. Im imaging the typeUrl to never be empty in `Mapper.GetKey`

jyotimahapatra: Also notice ADS is not a kind of resource https://github.com/envoyproxy/go-control-plane/blob/master/pkg/cache/resource.go#L36-L43

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,35 @@
...
+func NewMapper(config aggregationv1.KeyerConfiguration) Mapper {
+	return &mapper{
+		config: config,
+	}
+}
+
+// GetKey converts a request into an aggregated key
+func (mapper *mapper) GetKey(request v2.DiscoveryRequest, typeURL string) (string, error) {
+	if typeURL == "" {
+		return "", fmt.Errorf("Typeurl is empty")
```
jessicayuen: nit: typeURL is empty

internal/pkg/mapper/mapper.go
jyotimahapatra: Added more comments https://github.com/envoyproxy/xds-relay/pull/19/files#diff-6d8632086beef4726597650f47c0bec7R15-R17

internal/pkg/mapper/mapper.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// Mapper defines the interface that Maps an incoming request to an aggregation key
+type Mapper interface {
+	// GetKey converts a request into an aggregated key
+	// Returns error if the regex parsing in the config fails to compile or match
+	// Returns error if the typeURL is empty. An empty typeURL signifies an ADS request.
+	// ref: envoyproxy/envoy/blob/d1a36f1ea24b38fc414d06ea29c5664f419066ef/api/envoy/api/v2/discovery.proto#L43-L46
+	// ref: envoyproxy/go-control-plane/blob/1152177914f2ec0037411f65c56d9beae526870a/pkg/server/server.go#L305-L310
+	// The go-control-plane will always traslate typeUrl implicitly to one of the resource typeUrls.
```
jessicayuen: typo: s/traslate/translate, s/typeUrl/typeURL

jyotimahapatra: done

jyotimahapatra:
@jessicayuen i think i covered all comments now.

jyotimahapatra:
@LisaLudique @eapolinario if you have any other comments, feel free to dump here and I will take care of them in next set of PRs

PR #20, opened 2020-03-13
Mapper.GetKey impl: any match to string result

jyotimahapatra:
This PR implements the AnyMatch MatchPredicate and StringFragment Result.

Even though each test case has a single `TableEntry` its defined in a `var` so that more tests can be added without editing the test case itself in the subsequent PRs.

### DIFF
--- a/internal/app/mapper/mapper.go
+++ b/internal/app/mapper/mapper.go
@@ -2,10 +2,14 @@ package mapper
 
 import (
 	"fmt"
+	"strings"
 
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	aggregationv1 "github.com/envoyproxy/xds-relay/pkg/api/aggregation/v1"
 )
+
+type matchPredicate = aggregationv1.MatchPredicate
+type rule = aggregationv1.KeyerConfiguration_Fragment_Rule
 
 // Mapper defines the interface that Maps an incoming request to an aggregation key
 type Mapper interface {
@@ -14,7 +18,7 @@ type Mapper interface {
 	// Returns error if the typeURL is empty. An empty typeURL signifies an ADS request.
 	// ref: envoyproxy/envoy/blob/d1a36f1ea24b38fc414d06ea29c5664f419066ef/api/envoy/api/v2/discovery.proto#L43-L46
 	// ref: envoyproxy/go-control-plane/blob/1152177914f2ec0037411f65c56d9beae526870a/pkg/server/server.go#L305-L310
-	// The go-control-plane will always translate typeURL implicitly to one of the resource typeURL.
+	// The go-control-plane will always translate DiscoveryRequest typeURL to one of the resource typeURL.
 	GetKey(request v2.DiscoveryRequest, typeURL string) (string, error)
 }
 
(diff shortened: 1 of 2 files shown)


jyotimahapatra:
PTAL @LisaLudique @eapolinario @jessicayuen

internal/app/mapper/mapper.go
```diff
@@ -34,5 +38,44 @@ func (mapper *mapper) GetKey(request v2.DiscoveryRequest, typeURL string) (strin
...
+	}
+
+	if len(resultFragments) != 0 {
+		return strings.Join(resultFragments, "_"), nil
+	}
+
 	return "", fmt.Errorf("Cannot map the input to a key")
 }
+
+func isMatchPredicate(matchPredicate *matchPredicate) bool {
```
jessicayuen: shouldn't this be `isAnyMatch(...)`?
if it's to support the other future types would prefer to name this `isMatch(...)`

internal/app/mapper/mapper.go
```diff
@@ -34,5 +38,44 @@ func (mapper *mapper) GetKey(request v2.DiscoveryRequest, typeURL string) (strin
...
+	if matchPredicate.GetAnyMatch() {
+		return isAnyMatch(matchPredicate)
+	}
+
+	return false
+}
+
+func isAnyMatch(matchPredicate *matchPredicate) bool {
```
jessicayuen: Would just kill this func and `return isAnyMatch(matchPredicate)` in the above func. It will be the final else condition when the other match predicate types are supported.

internal/app/mapper/mapper.go
```diff
@@ -34,5 +38,44 @@ func (mapper *mapper) GetKey(request v2.DiscoveryRequest, typeURL string) (strin
...
+	return matchPredicate.GetAnyMatch()
+}
+
+func getResult(fragmentRule *rule) string {
+	stringFragment := fragmentRule.GetResult().GetStringFragment()
+	if stringFragment != "" {
+		return stringFragment
+	}
+
+	return ""
```
jessicayuen: Can just be `return stringFragment`

internal/app/mapper/mapper_test.go
```diff
@@ -45,9 +112,91 @@ var _ = Describe("GetKey", func() {
...
+				},
+			}
+			mapper := NewMapper(protoConfig)
+			key, err := mapper.GetKey(getDiscoveryRequest(), clusterTypeURL)
+			Expect(expectedKey).To(Equal(key))
+			Expect(err).Should(BeNil())
+		},
+		multiFragmentPositiveTests...)
+
+	DescribeTable("should be return error for non matching multiple fragments",
```
jessicayuen: `should be able to return ... `

internal/app/mapper/mapper.go
jyotimahapatra: makes sense.

internal/app/mapper/mapper.go
jyotimahapatra: More predicates are coming the next PR in this function. `isMatch` sounds good

jyotimahapatra:
Thanks @jessicayuen . Please do take another pass

PR #21, opened 2020-03-14
RequestTypeMatch predicate

jyotimahapatra:
This PR adds the [RequestTypeMatch](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L50-L54) predicate

### DIFF
--- a/internal/app/mapper/mapper_test.go
+++ b/internal/app/mapper/mapper_test.go
@@ -18,13 +18,14 @@ type MatchPredicate = aggregationv1.MatchPredicate
 type ResultPredicate = aggregationv1.ResultPredicate
 
 const (
-	clusterTypeURL = "type.googleapis.com/envoy.api.v2.Cluster"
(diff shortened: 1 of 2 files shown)


jyotimahapatra:

jyotimahapatra:

internal/app/mapper/mapper_test.go
```diff
@@ -34,7 +35,25 @@ var postivetests = []TableEntry{
 			getAnyMatch(true),
 			getResultStringFragment(),
 			clusterTypeURL,
-			stringfragment,
+			stringFragment,
+		},
+	},
+	{
+		Description: "RequestTypeMatch Matches with a single typeurl",
```
jessicayuen: Do we need to add a test case for ordering of the type url? (second match vs first)

jyotimahapatra: good idea. Done

Issue #6, opened 2020-02-25
Implement .yaml to Aggregation proto converter


jessicayuen:
Blocked by #4 

Users should be able to bootstrap their initial configuration using `.yaml`. Define the transformation rules necessary to convert their bootstrap configuration into the proto defined in #4 .

PR #22, opened 2020-03-16
YAML to proto converter

eapolinario:
This PR adds support for reading yaml and converting to one of the proto messages defined in the [api](https://github.com/envoyproxy/xds-relay/tree/master/api/protos/aggregation/v1).

We convert from yaml to proto via an intermediary JSON representation. Note that we are using the [new version](https://blog.golang.org/a-new-go-api-for-protocol-buffers) of the golang API for protocol buffers. We're leveraging the [protojson](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson?tab=doc) package to handle the unmarshaling from JSON to the actual protos, more specifically so that `oneof` fields are handled correctly. It's also worth pointing out that the generated code for the protos is actually readable in this new version of the API.

Of note I also fixed a mistake in the initial definition of the protos, more specifically, how we were defining a conjunction of `ResultPredicate` was convoluted. I took the opportunity to rename the field from `RepeatedResultPredicate` to `AndResult`. 

Issue: https://github.com/envoyproxy/xds-relay/issues/6


internal/pkg/util/yamlprotoconverter_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+var _ = Describe("Yamlprotoconverter", func() {
+	DescribeTable("should be able to convert from yaml to proto",
+		func(yml string, protoToUnmarshal proto.Message, expectedProto proto.Message) {
+			err := FromYAMLToProto(yml, protoToUnmarshal)
+			Expect(err).To(BeNil())
+			Expect(proto.Equal(protoToUnmarshal, expectedProto)).To(Equal(true))
+		},
+		positiveTests...)
+})
```
jyotimahapatra: maybe add a negative test too?

internal/pkg/util/yamlprotoconverter_test.go
```diff
@@ -0,0 +1,214 @@
...
+				},
+			},
+		},
+	},
+}
+
+var _ = Describe("Yamlprotoconverter", func() {
+	DescribeTable("should be able to convert from yaml to proto",
+		func(yml string, protoToUnmarshal proto.Message, expectedProto proto.Message) {
+			err := FromYAMLToProto(yml, protoToUnmarshal)
```
jyotimahapatra: should `protoToUnmarshal` and `expectedProto` be always `KeyerConfiguration` ?

go.mod
```diff
@@ -5,12 +5,15 @@ go 1.14
...
+	github.com/golang/protobuf v1.4.0-rc.4
 	github.com/iancoleman/strcase v0.0.0-20191112232945-16388991a334 // indirect
 	github.com/lyft/protoc-gen-star v0.4.14 // indirect
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
 	github.com/spf13/afero v1.2.2 // indirect
+	github.com/stretchr/testify v1.4.0 // indirect
 	go.uber.org/zap v1.14.0
 	google.golang.org/grpc v1.27.1
+	google.golang.org/protobuf v1.20.1
```
jyotimahapatra: This is great!

internal/pkg/util/yamlprotoconverter_test.go
eapolinario: sure, it's in the TODO in the description. I'm going to add a couple of negative tests too. (I have to say I'm really enjoying this BDD-style of testing...)

internal/pkg/util/yamlprotoconverter_test.go
eapolinario: actually, no, the only requirement in the tests is that the types match. There might be a way to get an empty copy of a proto object, but I couldn't figure it out. (I also left it as a TODO item).

internal/pkg/util/yamlprotoconverter.go
```diff
@@ -0,0 +1,19 @@
+package yamlprotoconverter
```
jessicayuen: nit: how about `yamlproto.go`? in the spirit of `google.golang.org/protobuf/encoding/protojson`

internal/pkg/util/yamlprotoconverter.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"github.com/ghodss/yaml"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+func FromYAMLToProto(yml string, pb proto.Message) error {
```
jessicayuen: I know this is pretty self explanatory, but can you document evert exported func? Not quite sure why golint isn't picking up on this.

internal/pkg/util/yamlprotoconverter_test.go
```diff
@@ -0,0 +1,271 @@
+package yamlprotoconverter_test
+
+import (
+	. "github.com/envoyproxy/xds-relay/internal/pkg/util"
```
jessicayuen: shouldn't need this import

internal/pkg/util/yamlprotoconverter_test.go
```diff
@@ -0,0 +1,271 @@
...
+	aggregationv1 "github.com/envoyproxy/xds-relay/pkg/api/aggregation/v1"
+	. "github.com/onsi/ginkgo"
```
jessicayuen: nit: can you separate the xds-relay imports from the 3rd party libs?

internal/pkg/util/yamlprotoconverter_test.go
```diff
@@ -0,0 +1,271 @@
...
+					StringFragment: "abc",
+				},
+			},
+		},
+	},
+	{
+		Description: "test result predicate containing resource_names_fragment",
+		Parameters: []interface{}{
+			`
+resource_names_fragment:
```
jessicayuen: I think it would be cleaner to put these yaml definitions in a `testdata` fixtures directory,
and rather than `[]interface{}` make `Parameters` type
```
[]struct {
    yamlPath string
    proto []interface{}
}
```
(or even making `proto` more strict in the above example to `ResultPredicate` and `KeyerConfiguration`

internal/pkg/util/yamlprotoconverter.go
eapolinario: good idea!

internal/pkg/util/yamlprotoconverter_test.go
eapolinario: we need this to have access to `FromYAMLToProto`, no?

eapolinario: nvm, the package name for tests was wrong.

internal/pkg/util/yamlproto_test.go
```diff
@@ -0,0 +1,213 @@
...
+			err = FromYAMLToProto(string(ymlBytes), protoToUnmarshal)
+			Expect(err).To(BeNil())
+			Expect(proto.Equal(protoToUnmarshal, expectedProto)).To(Equal(true))
+		},
+		positiveTests...)
+
+	DescribeTable("should not be able to convert from yaml to proto",
+		func(yml string, protoToUnmarshal proto.Message) {
+			err := FromYAMLToProto(yml, protoToUnmarshal)
+			Expect(err).To(HaveOccurred())
```
jyotimahapatra: can the error check be a bit more descriptive?

internal/pkg/util/yamlproto_test.go
```diff
@@ -0,0 +1,213 @@
...
+			protoToUnmarshal := proto.Clone(expectedProto)
+			proto.Reset(protoToUnmarshal)
+			err = FromYAMLToProto(string(ymlBytes), protoToUnmarshal)
+			Expect(err).To(BeNil())
+			Expect(proto.Equal(protoToUnmarshal, expectedProto)).To(Equal(true))
+		},
+		positiveTests...)
+
+	DescribeTable("should not be able to convert from yaml to proto",
+		func(yml string, protoToUnmarshal proto.Message) {
```
jyotimahapatra: may be this can be removed from param, since its always `KeyerConfiguration`?

jyotimahapatra:
LGTM modulo 2 comments.

internal/pkg/util/yamlproto_test.go
eapolinario: I'd rather add a few more negative tests for the inner protos, that way when we expand the list of protos we can build on these tests. How does that sound?

jyotimahapatra: Sounds good.

internal/pkg/util/yamlproto.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"github.com/ghodss/yaml"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+// FromYAMLToProto unmarshals a YAML string into a proto message.
+func FromYAMLToProto(yml string, pb proto.Message) error {
+	js, err := yaml.YAMLToJSON([]byte(yml))
```
jyotimahapatra: sorry for bringing this up late, but i think this is a good place to call the `.Validate()` on KeyerConfiguration proto. WDYT?

eapolinario: this is a good idea, but I don't think we can do this generally (i.e. there is no `Validate` method in `proto.Message`), which means that we're going to have specialized functions per proto type.

eapolinario: Initially I thought that putting the burden on the callers would be a reasonable tradeoff, but on second thought it's better to sanitize protos as early as possible.

jyotimahapatra: may be FromYAMLToProto should take a `KeyerConfiguration` proto so that we can call Validate() ? Is there another place you think the validation can be done?

eapolinario: I remember @jessicayuen mentioning that we should be able to unmarshal other types, but I might be misremembering. In any case, having a specific function for `KeyerConfiguration` is future-proof enough.

jyotimahapatra: Yeah. I think in xds-relay the input file is going to be a `KeyerConfiguration` always. So might be better to solve for that type explicitly.

internal/pkg/util/yamlprotoconverter_test.go
eapolinario: @jessicayuen , we cannot "upcast" to a [slice of interface](https://golang.org/doc/faq#convert_slice_of_interface).

internal/pkg/util/testdata/error_in_second_rule.yaml
```diff
@@ -0,0 +1,17 @@
+fragments:
+  - ruless:
```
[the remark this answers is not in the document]
eapolinario: this is actually a [test case](https://github.com/envoyproxy/xds-relay/blob/31ec4eb44e0316040e514dcd4b824ae693ef5eeb/internal/pkg/util/yamlproto_test.go#L187-L193) to show that a typo is caught during unmarshaling.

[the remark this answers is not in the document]
jessicayuen: ah I think I was confused by the title of the test `error_in_second_rule.yaml`

PR #23, opened 2020-03-16
RequestNodeMatch predicate

jyotimahapatra:
This PR adds the [RequestNodeMatch](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L57-L74) predicate

### DIFF
--- a/internal/app/mapper/mapper.go
+++ b/internal/app/mapper/mapper.go
@@ -69,7 +69,7 @@ func (mapper *mapper) GetKey(request v2.DiscoveryRequest, typeURL string) (strin
 }
 
 func isMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node) (bool, error) {
-	result, err := isRequestTypeMatch(matchPredicate, typeURL)
+	result, err := isNodeMatch(matchPredicate, node)
 	if err != nil {
 		return false, err
 	}
@@ -77,23 +77,7 @@ func isMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node) (b
 		return result, nil
 	}
 
-	result, err = isNodeMatch(matchPredicate, node)
-	if err != nil {
-		return false, err
-	}
-	if result {
-		return result, nil
-	}
-
-	result, err = isAnyMatch(matchPredicate)
-	if err != nil {
-		return false, err
-	}
-	if result {
-		return result, nil
-	}
-
-	return false, nil
+	return isRequestTypeMatch(matchPredicate, typeURL) || isAnyMatch(matchPredicate), nil
 }
 
 func isNodeMatch(matchPredicate *matchPredicate, node *core.Node) (bool, error) {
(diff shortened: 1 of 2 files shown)


jyotimahapatra:
PTAL @LisaLudique @eapolinario @jessicayuen

jessicayuen:
@jyotimahapatra can you merge in #24 ?

jyotimahapatra:
> @jyotimahapatra can you merge in #24 ?

done

internal/app/mapper/mapper.go
```diff
@@ -63,8 +68,37 @@ func (mapper *mapper) GetKey(request v2.DiscoveryRequest) (string, error) {
...
+func isMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node) (bool, error) {
+	result, err := isNodeMatch(matchPredicate, node)
+	if err != nil {
+		return false, err
+	}
+	if result {
+		return result, nil
+	}
+
+	return isRequestTypeMatch(matchPredicate, typeURL) || isAnyMatch(matchPredicate), nil
```
jessicayuen:
```suggestion
	isNodeMatch, err := isNodeMatch(matchPredicate, node)
	if err != nil {
		return false, err
	}

	return isNodeMatch || isRequestTypeMatch(matchPredicate, typeURL) || isAnyMatch(matchPredicate), nil
```

internal/app/mapper/mapper.go
```diff
@@ -89,3 +123,21 @@ func getResult(fragmentRule *rule) string {
 	stringFragment := fragmentRule.GetResult().GetStringFragment()
 	return stringFragment
 }
+
+func compare(requestNodeMatch *aggregationv1.MatchPredicate_RequestNodeMatch, nodeValue string) (bool, error) {
+	exactMatch := requestNodeMatch.GetExactMatch()
+	if exactMatch != "" {
```
jessicayuen: We would get a false negative if the nodeValue is empty right? Perhaps that is desired, but something to consider. Regardless would add a test case for this scenario.

internal/app/mapper/mapper_test.go
```diff
@@ -65,6 +70,96 @@ var postivetests = []TableEntry{
...
+			getRequestNodeRegexMatch(nodeIDField, "n....d"),
+			getResultStringFragment(),
+			clusterTypeURL,
+			stringFragment,
+		},
+	},
+	{
+		Description: "RequestNodeMatch with node cluster regex match",
+		Parameters: []interface{}{
+			getRequestNodeRegexMatch(nodeClusterField, "c.....r"),
```
jessicayuen: Can we try some other regex variations in a few of these tests? ex: `c*`

jyotimahapatra: sounds good. Added some more.

internal/app/mapper/mapper.go
jyotimahapatra: i think that is desired. Added some tests now.

jyotimahapatra:
PTAL @jessicayuen

jyotimahapatra:
@jessicayuen Thanks..I will work on the next ones once this gets merged.

PR #24, opened 2020-03-16
remove typeUrl dependency in mapper

jyotimahapatra:
DiscoveryRequest will always contain typeUrl
ADS will contain typeUrl https://github.com/envoyproxy/envoy/blob/master/api/envoy/api/v2/discovery.proto#L46
Implicit xds requests will have typeUrl set because go-control-plane mutates the DiscoveryRequest
ref: https://github.com/envoyproxy/go-control-plane/blob/master/pkg/server/server.go#L310


jyotimahapatra:

PR #25, opened 2020-03-17
Cache API Implementation using Ristretto

LisaLudique:
This PR implements the basic methods defined in the cache interface using the [Ristretto caching library](https://github.com/dgraph-io/ristretto). The cache is configured based on the suggestions in https://github.com/dgraph-io/ristretto#config - note that when setting a value, the size in bytes of the value is used as a "cost" measurement.

#9


internal/app/cache/cache.go
```diff
@@ -17,3 +22,104 @@ type Cache interface {
...
+	expireSeconds time.Duration
+}
+
+type resource struct {
+	resp       *envoy_api_v2.DiscoveryResponse
+	watches    []*envoy_api_v2.DiscoveryRequest
+	streamOpen bool
+}
+
+func newCache(cacheSizeBytes int64, expireSeconds int) (*cache, error) {
```
jessicayuen: Shouldn't this function be exported, and return the interface `Cache`?

internal/app/cache/cache.go
```diff
@@ -17,3 +22,104 @@ type Cache interface {
...
+	config := ristretto.Config{
+		NumCounters: 10 * cacheSizeBytes,
+		MaxCost:     cacheSizeBytes,
+		BufferItems: 64,
```
jessicayuen: Can you add a comment for what these signify?

internal/app/cache/cache.go
```diff
@@ -17,3 +22,104 @@ type Cache interface {
...
+	resource, ok := value.(resource)
+	if !ok {
+		return nil, fmt.Errorf("Unable to cast cache value to type resource for key: %s", key)
+	}
+	return resource.resp, nil
+}
+
+func (c *cache) SetResponse(key string, resp envoy_api_v2.DiscoveryResponse) ([]*envoy_api_v2.DiscoveryRequest, error) {
+	value, found := c.cache.Get(key)
+	if !found {
```
jessicayuen: This is an interesting scenario, because we're implying that a response can come from the origin server with no open watches. I suppose this could be the case if it so happened that the cache TTL expired when the response arrived.

internal/app/cache/cache.go
```diff
@@ -17,3 +22,104 @@ type Cache interface {
...
+			return nil, fmt.Errorf("Unable to set value for key: %s", key)
+		}
+		return nil, nil
+	}
+	resource, ok := value.(resource)
+	if !ok {
+		return nil, fmt.Errorf("Unable to cast cache value to type resource for key: %s", key)
+	}
+	resource.resp = &resp
+	resource.watches = nil
```
jessicayuen: Hmm I think we agreed to keep the watches open, no?. Sorry, the spec is a bit dated in this regards.

internal/app/cache/cache.go
```diff
@@ -17,3 +22,104 @@ type Cache interface {
...
+		return nil, fmt.Errorf("No value found for key: %s", key)
+	}
+	resource, ok := value.(resource)
+	if !ok {
+		return nil, fmt.Errorf("Unable to cast cache value to type resource for key: %s", key)
+	}
+	return resource.resp, nil
+}
+
+func (c *cache) SetResponse(key string, resp envoy_api_v2.DiscoveryResponse) ([]*envoy_api_v2.DiscoveryRequest, error) {
```
jessicayuen: Not something we have to tackle now, but eventually we're going to either need to:
1. Add logic here responsible for directly notifying watches or,
2. Change the interface here to add observers that the response has changed, so that the orchestrator can subsequently notify watches.

internal/app/cache/cache.go
```diff
@@ -17,3 +22,104 @@ type Cache interface {
...
+			return false, fmt.Errorf("Unable to set value for key: %s", key)
+		}
+		return true, nil
+	}
+	resource, ok := value.(resource)
+	if !ok {
+		return false, fmt.Errorf("Unable to cast cache value to type resource for key: %s", key)
+	}
+	resource.watches = append(resource.watches, &req)
+	resource.streamOpen = true
```
jessicayuen: Likewise, we'll need to wire up something here to guarantee a stream has been opened against the origin server before setting this flag. This is probably a non-trivial component that requires that requires the client and back-off/retries, so I'm ok with just adding a TODO.

internal/app/cache/cache.go
LisaLudique: Yes, for SetResponse (and AddWatch) I had initially made the behavior such that if no entry exists a new value is created and added to the cache anyway. Then if a new call to Fetch is made the latest response could still be sent over.

internal/app/cache/cache.go
LisaLudique: I read the docs again more closely, and NumCounters should actually be 10x the number of unique items in a full cache and not 10x the MaxCost necessarily. I'm leaning towards making this value configurable.

internal/app/cache/cache.go
```diff
@@ -17,3 +22,112 @@ type Cache interface {
...
 }
+
+type cache struct {
+	cache         *ristretto.Cache
+	expireSeconds time.Duration
+}
+
+type resource struct {
+	resp       *envoy_api_v2.DiscoveryResponse
+	watches    []*envoy_api_v2.DiscoveryRequest
```
jyotimahapatra: I think the watches are go channels

LisaLudique: Are we replacing watchers with response channels (https://envoyproxy.slack.com/archives/CTULFMN91/p1584409594031100) instead of DiscoveryRequests? This seems like a deviation from the cache interface, unless there's a way to get the DiscoveryRequest from its channel.

internal/app/cache/cache.go
```diff
@@ -17,3 +22,116 @@ type Cache interface {
 	// AddWatch adds the watch to the cache and returns whether a stream is open.
 	AddWatch(key string, req envoy_api_v2.DiscoveryRequest) (bool, error)
 }
+
+type cache struct {
+	cache         *ristretto.Cache
+	expireSeconds time.Duration
```
jyotimahapatra: may be drop `Seconds` because the type is `time.Duration`?

internal/app/cache/cache.go
```diff
@@ -17,3 +22,116 @@ type Cache interface {
...
+		// and 10x the number of unique items in the cache when full (note that this is different from MaxCost).
+		NumCounters: numCounters,
+		// MaxCost denotes the max size in bytes and determines how eviction decisions are made.
+		// For consistency, the cost passed into SetWithTTL represents the size of the value in bytes.
+		MaxCost: cacheSizeBytes,
+		// BufferItems is the size of the Get buffers. The recommended value is 64.
+		BufferItems: 64,
+		// OnEvict is called for each eviction and closes the stream if a key is removed (e.g. expiry due to TTL).
+		OnEvict: func(key, conflict uint64, value interface{}, cost int64) {
+			// TODO: Add logic to guarantee that the corresponding stream is closed.
```
jyotimahapatra: the func could be an input to `NewCache`

internal/app/cache/cache.go
```diff
@@ -17,3 +22,116 @@ type Cache interface {
...
+	set := c.cache.SetWithTTL(key, resource, int64(cost), c.expireSeconds)
+	// TODO: Add logic that allows for notifying of watches.
+	if !set {
+		return resource.watches, fmt.Errorf("Unable to set value for key: %s", key)
+	}
+	return resource.watches, nil
+}
+
+func (c *cache) AddWatch(key string, req envoy_api_v2.DiscoveryRequest) (bool, error) {
+	value, found := c.cache.Get(key)
```
jyotimahapatra: do you think we will need locks anywhere?

internal/app/cache/cache.go
jyotimahapatra: sorry, my point was to call it something else, instead of watches.

LisaLudique: Gotcha. I think the spec had actually called them watchers instead of watches (which I'd missed in my initial PR) but since the terms are used interchangeably and might lead to confusion, I'm just going to rename this to requests.

internal/app/cache/cache.go
LisaLudique: The unit is set to seconds in L60. I'm not sure how else to denote it in terms of seconds to make it clear for the user.

internal/app/cache/cache.go
LisaLudique: With the current implementation I think each method will need to be protected.

internal/app/cache/cache.go
```diff
@@ -11,9 +17,130 @@ type Cache interface {
...
+	expireSeconds time.Duration
+}
+
+type resource struct {
+	resp       *envoy_api_v2.DiscoveryResponse
+	requests   []*envoy_api_v2.DiscoveryRequest
+	streamOpen bool
+}
+
+func NewCache(numCounters int64, cacheSizeBytes int64, expireSeconds int, onEvict func(key, conflict uint64,
```
jessicayuen: thoughts on pulling out `onEvict` into its own type? Would make easier to document what some of the variables (key, conflict, value, cost) mean.

```
// onEvict ...
type onEvict func(uint64, uint64, interface{}, int64)
```

internal/app/cache/cache.go
```diff
@@ -11,9 +17,130 @@ type Cache interface {
...
+
+func (c *cache) Exists(key string) bool {
+	c.cacheMu.RLock()
+	_, found := c.cache.Get(key)
+	c.cacheMu.RUnlock()
+	return found
+}
+
+func (c *cache) Fetch(key string) (*envoy_api_v2.DiscoveryResponse, error) {
+	c.cacheMu.RLock()
```
jessicayuen: What's the reasoning behind locking on reads?

LisaLudique: I was concerned about reads to the cache kicking off multiple writes if AddRequest/SetResponse hadn't finished, but I don't think the locking on reads is needed for Exists/Fetch.

internal/app/cache/cache.go
LisaLudique: Done. The documentation is a bit lacking on the first two parameters, but tldr we lose what the original value of the key is when this is called since we only passed in the hashed key.

internal/app/cache/cache.go
jyotimahapatra: `SetWithTTL` takes a `time.Duration`. https://github.com/dgraph-io/ristretto/blob/master/cache.go#L201
I think it doesn't need seconds.

internal/app/cache/cache.go
```diff
@@ -11,9 +17,128 @@ type Cache interface {
...
+		return nil, err
+	}
+	return &cache{
+		cache: newCache,
+		ttl:   ttl,
+	}, nil
+}
+
+func (c *cache) Exists(key string) bool {
+	_, found := c.cache.Get(key)
```
jessicayuen: We should probably check for whether the response exists for the key rather than if any value exists.

Originally I imagined that the orchestrator would call this prior to `Fetch`, but thinking more we could probably remove this func altogether in favor of just nil checking on `Fetch`.

Can you think of other use cases where we might want to call this `Exists`?

jessicayuen: cc @jyotimahapatra

jyotimahapatra: SGTM

jyotimahapatra: Do you mean the use case will use nil checking on https://github.com/envoyproxy/xds-relay/pull/25/files#diff-daf2807d54a5369eb2a29014b0d051cfR81 ?

internal/app/cache/cache.go
jyotimahapatra: sounds good

internal/app/cache/cache_test.go
```diff
@@ -0,0 +1,136 @@
...
+	Canary:      false,
+	TypeUrl:     "typeURL_A",
+	Nonce:       "nonce_A",
+	ControlPlane: &core.ControlPlane{
+		Identifier: "identifier_A",
+	},
+}
+
+func TestExists_EmptyCache(t *testing.T) {
+	cache, err := NewCache(10, 1048576, time.Second*60, onEvict)
```
jyotimahapatra: I was thinking more about this..Do we use two different testing patterns or just the BDD pattern everywhere?
@jessicayuen @eapolinario @LisaLudique What is your opinion?

internal/app/cache/cache_test.go
```diff
@@ -0,0 +1,136 @@
...
+	assert.Equal(t, testRequest, *requests[0])
+	assert.Equal(t, testRequest, *requests[1])
+
+	response, err := cache.Fetch(testKeyA)
+	assert.NoError(t, err)
+	assert.Equal(t, testResponse, *response)
+}
+
+func TestTTL(t *testing.T) {
+	cache, err := NewCache(10, 1048576, time.Second*1, onEvict)
```
jyotimahapatra: can the ttl be smaller for the testcase?

internal/app/cache/cache_test.go
```diff
@@ -0,0 +1,136 @@
...
+	_, err = cache.AddRequest(testKeyA, testRequest)
+	assert.NoError(t, err)
+	time.Sleep(1 * time.Millisecond)
+	assert.True(t, cache.Exists(testKeyA))
+	time.Sleep(1 * time.Second)
+	assert.False(t, cache.Exists(testKeyA))
+}
+
+func TestMemoryOverflow(t *testing.T) {
+	cache, err := NewCache(10, 40, time.Second*60, onEvict)
```
jyotimahapatra: can we add a test for onEvict too?

internal/app/cache/cache.go
LisaLudique: Agreed, you can just check if the return value from Fetch is nil if that was the intent.

LisaLudique: FWIW it was useful for test validation

internal/app/cache/cache_test.go
eapolinario: I don't think we should be married to BDD necessarily, i.e. use it when it makes sense, otherwise, regular xUnit-type tests can (and should) be used.

eapolinario:
More of a general question (from a go n00b): is there any way we can mock calls to `time.Sleep`?

internal/app/cache/cache_test.go
jyotimahapatra: sounds good

internal/app/cache/cache_test.go
```diff
@@ -0,0 +1,136 @@
...
+	assert.NoError(t, err)
+	_, err = cache.AddRequest(testKeyA, testRequest)
+	assert.NoError(t, err)
+	time.Sleep(1 * time.Millisecond)
```
jessicayuen: Echoing what @eapolinario said, I would avoid putting these sleeps in the tests. Do we need them right now?

LisaLudique: https://github.com/envoyproxy/xds-relay/blob/e385cbec086fdc08ac8c46bfcefcc4ff9fb993a8/internal/app/cache/cache_test.go#L84-L107 demonstrates the ristretto cache behavior, in which without the sleep between AddRequest calls (https://github.com/envoyproxy/xds-relay/blob/e385cbec086fdc08ac8c46bfcefcc4ff9fb993a8/internal/app/cache/cache_test.go#L93), only 1 request will be associated with that key (https://github.com/envoyproxy/xds-relay/blob/e385cbec086fdc08ac8c46bfcefcc4ff9fb993a8/internal/app/cache/cache_test.go#L100 would fail). I'm not sure why the cache drops the request, even with the lock.

jessicayuen:
Closing this one to clear my visual queue of PRs to review. :) Feel free to re-open if you prefer.

PR #26, opened 2020-03-17
And and Or predicates

jyotimahapatra:
Adding the [And and Or](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L88-L92) predicates

### DIFF
--- a/internal/app/mapper/mapper.go
+++ b/internal/app/mapper/mapper.go
@@ -73,7 +73,6 @@ func isMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node) (b
 	if err != nil {
 		return false, err
 	}
-
 	if isNodeMatch {
 		return true, nil
 	}
--- a/internal/app/mapper/mapper.go
+++ b/internal/app/mapper/mapper.go
@@ -78,7 +78,7 @@ func isMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node) (b
 		return isNodeMatch, nil
 	}
 
-	isAndMatch, err := isAndMatch(matchPredicate, node, typeURL)
+	isAndMatch, err := isAndMatch(matchPredicate, typeURL, node)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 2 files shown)


jyotimahapatra:

internal/app/mapper/mapper.go
```diff
@@ -74,7 +74,27 @@ func isMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node) (b
 		return false, err
 	}
 
-	return isNodeMatch || isRequestTypeMatch(matchPredicate, typeURL) || isAnyMatch(matchPredicate), nil
+	if isNodeMatch {
```
jyotimahapatra: @jessicayuen I know this function has seen the most style comments. Any ideas how to handle this mess now?

jyotimahapatra:
PTAL @jessicayuen

internal/app/mapper/mapper.go
eapolinario: Working with protos is a bit clunky, but can we find out which type we're going to handle and have a switch instead of this sequence of if-else blocks?

jyotimahapatra: @eapolinario can you tell bit more about `which type we're going to handle` ?

jessicayuen: @jyotimahapatra I think this is ok. I can't see us getting around this if we want to return early on result match or error scenarios.

internal/app/mapper/mapper.go
```diff
@@ -74,7 +74,27 @@ func isMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node) (b
 		return false, err
 	}
```
jessicayuen: super nit: delete this line (to be consistent with the others)

internal/app/mapper/mapper.go
```diff
@@ -116,6 +136,42 @@ func isAnyMatch(matchPredicate *matchPredicate) bool {
...
+
+	for _, rule := range matchset.GetRules() {
+		isMatch, err := isMatch(rule, typeURL, node)
+		if err != nil {
+			return false, err
+		}
+		if !isMatch {
+			return false, nil
+		}
+	}
```
jessicayuen: This is very clean, I liked how you handled both these predicate scenarios.

jyotimahapatra: done

PR #27, opened 2020-03-18
NotMatch predicate

jyotimahapatra:
This implements [not_match](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L96) predicate


jyotimahapatra:

PR #28, opened 2020-03-18
Add interface for upstream control plane client

jyotimahapatra:
This PR adds the interface for the component that interacts with the upstream control plane.


jyotimahapatra:
PTAL @eapolinario @LisaLudique @jessicayuen

internal/app/stream/multiplexer.go
```diff
@@ -0,0 +1,34 @@
+package multiplexer
```
jyotimahapatra: do you prefer this to live in the server package instead?

jessicayuen: No, I think what you have is good. `server` was intended for the xDS relay gRPC server implementation. Thoughts on `app/upstream/multiplexer.go` vs `/app/stream/...` though?

jyotimahapatra: yeah `upstream` sounds much better.

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,34 @@
...
+	QueueRequest(context.Context, chan *v2.DiscoveryRequest, chan *v2.DiscoveryResponse)
+}
+
+type multiplexer struct {
+	conn *grpc.ClientConn
+}
+
+// NewMux creates an instance based on the typeUrl of the resource
+// A new instance of Multiplexer is recommended per xds type
+// Eg. For eds requests of different services, create an instance each.
```
eapolinario: grammar nits: 
- `s/Eg./e.g.,/`
- lowercase `For`
- capitalize `eds`

How about invert the sentence order? More specifically, I'm suggesting we rephrase this third line as:
```suggestion
// e.g., create an instance for each EDS request of different services
```

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,34 @@
...
+type Multiplexer interface {
+	// QueueRequest starts a stream with the upstream control plane
+	QueueRequest(context.Context, chan *v2.DiscoveryRequest, chan *v2.DiscoveryResponse)
+}
+
+type multiplexer struct {
+	conn *grpc.ClientConn
+}
+
+// NewMux creates an instance based on the typeUrl of the resource
```
eapolinario: grammar nit: can you and a period at the end of this line?

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	"google.golang.org/grpc"
+)
+
+// Multiplexer handles the connections to the upstream control plane
+type Multiplexer interface {
+	// QueueRequest starts a stream with the upstream control plane
+	QueueRequest(context.Context, chan *v2.DiscoveryRequest, chan *v2.DiscoveryResponse)
```
eapolinario: Do calls to this function always succeed?

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,34 @@
+package multiplexer
+
+import (
+	"context"
+
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	"google.golang.org/grpc"
+)
+
+// Multiplexer handles the connections to the upstream control plane
```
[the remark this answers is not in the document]
jyotimahapatra: makes sense.done.

internal/app/upstream/multiplexer.go
jyotimahapatra: Yes..The function will do async calls and retries internally. I'm still reading up on how to handle errors and bubble them up when using channels.

jyotimahapatra: found that using an error pipeline is a wellknown pattern , [reading](https://blog.golang.org/pipelines),
[followup reading](https://medium.com/statuscode/pipeline-patterns-in-go-a37bb3a7e61d)

internal/app/upstream/multiplexer.go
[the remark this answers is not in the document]
jessicayuen: Sorry to be even more picky. I think it's important we establish a consistent glossary. Matt chose to use the term `origin server` when setting up the repo and blog post, and I think it makes sense to go with that terminology: `Caching, aggregation, and relaying for xDS compliant clients and origin servers`

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,37 @@
...
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	"google.golang.org/grpc"
+)
+
+// Multiplexer handles the connections to the upstream management server
+type Multiplexer interface {
+	// QueueRequest starts a stream with the upstream management server
+	// All incoming discovery requests arrive on the request channel
+	// All responses go out through the response channel
+	// All irrecoverable errors go out through the error channel
```
jessicayuen: Hmm, can we get rid of the error channel for now? I think it creates a disconnect between the request/response that caused the error and the error itself.

What do you think of wrapping `v2.DiscoveryResponse` in a custom struct (or Proto) that defines the discovery response and an optional error? And have a channel of the custom object type instead?

jessicayuen:
@mattklein123 This is a core component, would appreciate an additional review.

internal/app/upstream/multiplexer.go
jyotimahapatra: I saw https://blog.golang.org/pipelines that using a separate error channel is a pattern. Also it will bee used for irrrecoverable errors. We definitely need a way to bubble up some errors to the caller. The caller could use that to cancel the context and shutdown running goroutines.

jyotimahapatra: I can remove the error altogether if you have a strong recommendation for it.

internal/app/upstream/multiplexer.go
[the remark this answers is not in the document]
jyotimahapatra: That makes sense, 
Instead of `Multiplexer handles the connections to the upstream management server` how does this sound?
`Multiplexer handles the connections to the origin server`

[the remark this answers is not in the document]
jyotimahapatra: done

internal/app/upstream/multiplexer.go
mattklein123: +1 to simplify for now and also for using a wrapped struct to carry per-request/response errors.

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,37 @@
...
+	"google.golang.org/grpc"
+)
+
+// Multiplexer handles the connections to the origin server
+type Multiplexer interface {
+	// QueueRequest starts a stream with the origin server
+	// All incoming discovery requests arrive on the request channel
+	// All responses go out through the response channel
+	// All irrecoverable errors go out through the error channel
+	QueueRequest(context.Context, chan *v2.DiscoveryRequest, chan *v2.DiscoveryResponse, chan error)
```
mattklein123: Does this need to return a handle or some way of cleaning up? How do things get removed from the multiplexer?

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	"google.golang.org/grpc"
+)
+
+// Multiplexer handles the connections to the origin server
+type Multiplexer interface {
+	// QueueRequest starts a stream with the origin server
+	// All incoming discovery requests arrive on the request channel
```
mattklein123: nit: s/incoming/outgoing, s/responses go out/responses come in below

internal/app/upstream/multiplexer.go
jyotimahapatra: QueueRequest now sends an error.

internal/app/upstream/multiplexer.go
jyotimahapatra: I have removed the error channel for now and wrapping it with the response.

internal/app/upstream/multiplexer.go
jyotimahapatra: The context.cancel should be able to perform the cleanups, it will be part of implementation.

jyotimahapatra:
@mattklein123 I have added some more comments. I drew the name inspiration from [envoy](https://github.com/envoyproxy/envoy/blob/master/source/common/config/grpc_mux_impl.cc). If the name is misleading we can call it `upstreamResponseManager` or some variant of it.

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	"google.golang.org/grpc"
+)
+
+// Multiplexer handles the requests and responses from the origin server.
+// The multiplexer handles each xds request on a separate stream,
+// e.g. 2 different cds requests happen on 2 separate streams.
+// It is the caller's responsibility to make sure there is one instance of Multipler per unique xds request.
```
mattklein123: typo "Multipler" ?

internal/app/upstream/multiplexer.go
```diff
@@ -0,0 +1,55 @@
...
+	QueueRequest(context.Context, chan *v2.DiscoveryRequest, chan *Response) error
+}
+
+type multiplexer struct {
+	conn *grpc.ClientConn
+}
+
+// Response struct is a holder for the result from a single request.
+// A request can result in a response from origin server or an error
+// One one of the fields is valid at any time. If the error is set, the response will be ignored.
```
mattklein123: s/One/Only

internal/app/upstream/multiplexer.go
mattklein123: I'm still a little confused what this interface is for. What is it multiplexing exactly if there is one required for each unique xDS request and the caller is responsible for figuring out uniqueness? Is this just an `XdsClient` interface?

jyotimahapatra: That is true and I've now renamed it to XdsClient.

internal/app/upstream/client.go
```diff
@@ -0,0 +1,57 @@
...
+
+type xdsClient struct {
+	conn *grpc.ClientConn
+}
+
+// Response struct is a holder for the result from a single request.
+// A request can result in a response from origin server or an error
+// Only one of the fields is valid at any time. If the error is set, the response will be ignored.
+type Response struct {
+	//nolint
```
jyotimahapatra: The `//nolint` is because golang complains that the field is unused. I will remove it in the the next PR while implementing.

internal/app/upstream/client.go
```diff
@@ -0,0 +1,57 @@
...
+	// All discovery requests to origin server arrive on the request channel
+	// All responses from the origin server are sent back through the response channel
+	// The caller is responsible for managing the lifetime of the channels.
+	// QueueRequest uses the retry and timeout configurations to make best effort to get the responses from origin server.
+	// If there's a new request in between retries, the retries are abandoned.
+	// The request and response happen asynchronously. Retries are scoped for sending messages to origin server.
+	// If the timeouts are exhausted, receive fails or a irrecoverable error occurs, the error is sent back to the caller.
+	// It is the caller's responsibility to send a new request from the last known DiscoveryRequest.
+	// Cancellation and cleanup operations will be based on cancellation of the context and closing of channels.
+	QueueRequest(context.Context, chan *v2.DiscoveryRequest, chan *Response) error
```
mattklein123: Thanks this makes more sense now. I think a better name for this method might be `CreateRequestResponseChannel` or similar? Since this doesn't actually queue anything?

jyotimahapatra: I like `xdsClient.Start()` or `xdsClient.Initialize()` better. Does one of these sound reasonable?

mattklein123: If this is a one shot function, why does it need to live on the interface at all? Can't it be part of the constructor for the thing below? Otherwise yes either Start or Initialize seems fine.

jyotimahapatra: The scenario we might need the function is when we initialize the constructor with args at a different place than where the function called. Since we don't have the impl structure yet, it's hard to imagine, but gives a reasonable way for other parts to Mock this interface and come up with other interfaces.
Although creating a higher order `func` will be ok for this purpose, i did not want to use it yet.

jyotimahapatra: renamed the function to `Start` for now.

internal/app/upstream/client.go
```diff
@@ -0,0 +1,57 @@
+package client
```
jessicayuen: Thoughts on naming this package `upstream` and the interface `Client`? So it's `upstream.Client` rather than `client.XdsClient`?

internal/app/upstream/client.go
jessicayuen: > The scenario we might need the function is when we initialize the constructor with args at a different place than where the function called.

I couldn't quite follow this, can you give me an example?

jessicayuen: I think that there should be a top level `NewClient` function that does what's in the interface currently, and the client interface would containing a contract for adding new streams. I'm on the same page as Matt that I don't think the one shot func should live in the interface.

```
package upstream

...

type Client interface {
    AddStream(*v2.DiscoveryRequest) error
}

NewClient(context.Context, chan *Response) (Client, error)
```

The implementation of AddStream would push it into the an internal request channel maintained by the client to handle retries and what not. The caller would still maintain the response channel to receive responses in a non-blocking manner.

What are your thoughts on this approach?

internal/app/upstream/client.go
jyotimahapatra: renamed the package to upstream

internal/app/upstream/client.go
jyotimahapatra: I made some changes in the interface and it is not a one shot function anymore.
I've also created a sample impl to brainstorm with. The `NewClient` will be a singleton instance and will not be per request. I am envisioning that the NewClient will hold a single grpc client connection and will manage closing the connection on shutdown.

Everything else remains the same. @jessicayuen about your proposed changes, i think we might be constrained with how grpc works for streams. Request and response have to be called on a single stream and are asynchronous and should always be called from a single goroutine each. Given that the client will be a singleton, do you think we still need to adopt the suggestion?

ref: https://github.com/envoyproxy/xds-relay/pull/36/files

jyotimahapatra:
@jessicayuen read for review now.

PR #30, opened 2020-03-19
LRU cache implementation

LisaLudique:
This PR implements the basic methods defined in the cache interface using the [groupcache/lru library](https://godoc.org/github.com/golang/groupcache/lru). In addition to eviction that occurs when a cache receives new items once full that traditionally happens with an LRU cache, users may configure a TTL after which an individual entry will expire.

#9


internal/app/cache/cache.go
```diff
@@ -1,19 +1,134 @@
...
+	return r.expirationTime.Before(time.Now())
+}
 
-	// AddWatch adds the watch to the cache and returns whether a stream is open.
-	AddWatch(key string, req envoy_api_v2.DiscoveryRequest) (bool, error)
+func (c *cache) getExpirationTime() time.Time {
+	if c.ttl > 0 {
+		return time.Now().Add(c.ttl)
+	}
+	return time.Time{}
```
jyotimahapatra: what is the default value? time.Now?

internal/app/cache/cache.go
```diff
@@ -1,19 +1,134 @@
...
+	if !found {
+		return nil, fmt.Errorf("no value found for key: %s", key)
+	}
+	resource, ok := value.(resource)
+	if !ok {
+		return nil, fmt.Errorf("unable to cast cache value to type resource for key: %s", key)
+	}
+	// Lazy eviction based on TTL occurs here. Fetch does not increase the lifespan of the key.
+	if resource.isExpired() {
+		c.cache.Remove(key)
```
jyotimahapatra: is onEvicted not implemented yet? Do we envision to implement it here?

internal/app/cache/cache_test.go
```diff
@@ -0,0 +1,154 @@
...
+	cache := NewCache(1, testOnEvict, time.Millisecond*10)
+
+	_, err := cache.SetResponse(testKeyA, testResponse)
+	assert.NoError(t, err)
+
+	response, err := cache.Fetch(testKeyA)
+	assert.NoError(t, err)
+	assert.Equal(t, testResponse, *response)
+
+	gomega.Eventually(func() (*v2.DiscoveryResponse, error) {
```
jyotimahapatra: since we are handling ttl on our own, can we try to eliminate `Eventually` by passing in a mock time?

internal/app/cache/cache.go
LisaLudique: The default value is time.Time's 0 value. From https://golang.org/pkg/time/#Time:

> The zero value of type Time is January 1, year 1, 00:00:00.000000000 UTC. As this time is unlikely to come up in practice, the IsZero method gives a simple way of detecting a time that has not been initialized explicitly.

`IsZero` is used to determine if expiration is disabled: https://github.com/envoyproxy/xds-relay/blob/a3ff4be3d6288f2982396770d38335059ad835c8/internal/app/cache/cache.go#L123

internal/app/cache/cache.go
LisaLudique: I left a TODO for onEvicted as 1) we are letting it be a parameter that is passed in and 2) for our purposes, onEvicted must guarantee the stream is closed, which will likely require more work outside the scope of this PR as per https://github.com/envoyproxy/xds-relay/pull/25#discussion_r393441180.

There's no need to explicitly call onEvicted since LRUCache invokes it upon [Remove](https://github.com/golang/groupcache/blob/8c9f03a8e57eb486e42badaed3fb287da51807ba/lru/lru.go#L111).

internal/app/cache/cache.go
```diff
@@ -1,19 +1,134 @@
...
+	return &cache{
+		cache: lru.Cache{
+			// Max number of cache entries before an item is evicted. Zero means no limit.
+			MaxEntries: maxEntries,
+			// OnEvict is called for each eviction and closes the stream if a key is removed
+			// (e.g. expiry due to TTL, too many entries).
+			OnEvicted: onEvicted,
+		},
+		// Duration before which an item is evicted for expiring. Zero means no expiration time.
+		ttl: ttl,
```
eapolinario: can you add a check to make sure `ttl` is non-negative?

internal/app/cache/cache.go
jyotimahapatra: makes sense. Is there a way to test that onEvicted was called?

LisaLudique: I asked in the slack thread - I am not sure how to test it was called.

internal/app/cache/cache.go
```diff
@@ -1,19 +1,137 @@
 package cache
```
jessicayuen: Could you add a package level comment for this component?

internal/app/cache/cache.go
```diff
@@ -1,19 +1,137 @@
...
+func NewCache(maxEntries int, onEvicted onEvictFunc, ttl time.Duration) (Cache, error) {
+	if ttl < 0 {
+		return nil, fmt.Errorf("ttl must be nonnegative but was set to %v", ttl)
+	}
+	return &cache{
+		cache: lru.Cache{
+			// Max number of cache entries before an item is evicted. Zero means no limit.
+			MaxEntries: maxEntries,
+			// OnEvict is called for each eviction and closes the stream if a key is removed
+			// (e.g. expiry due to TTL, too many entries).
```
jessicayuen: would remove `and closes the stream if a key is removed (e.g. expiry due to TTL, too many entries).` from here, since this is just the callback func.

internal/app/cache/cache.go
```diff
@@ -1,19 +1,137 @@
...
+			OnEvicted: onEvicted,
+		},
+		// Duration before which an item is evicted for expiring. Zero means no expiration time.
+		ttl: ttl,
+	}, nil
+}
+
+func (c *cache) Fetch(key string) (*v2.DiscoveryResponse, error) {
+	c.cacheMu.Lock()
+	defer c.cacheMu.Unlock()
```
jessicayuen: I thought we were going to get rid of the locks for reads?

internal/app/cache/cache.go
jessicayuen: onEvicted() is a callback func though right? Are you envisioning the caller to be responsible for shutting down the stream? If we're shutting down the stream as a part of the cache implementation, I would just add a TODO. It's separate from onEvict, the callback.

internal/app/cache/cache.go
```diff
@@ -1,19 +1,137 @@
...
+}
+
+func (c *cache) SetResponse(key string, resp v2.DiscoveryResponse) ([]*v2.DiscoveryRequest, error) {
+	c.cacheMu.Lock()
+	defer c.cacheMu.Unlock()
+	value, found := c.cache.Get(key)
+	if !found {
+		resource := resource{
+			resp:           &resp,
+			expirationTime: c.getExpirationTime(time.Now()),
```
jessicayuen: An alternative is to keep track of a `lastModified` value with is always `time.Now()` on set and reverse the logic in isExpired. We'd probably have more reads than writes though so what you have makes sense from a performance standpoint. lastModified may provide more information during debugging when we do a API cache dump. Don't have a strong preference, just some thoughts.

internal/app/cache/cache.go
LisaLudique: Since reads can potentially result in entry removal, I think Fetch needs a lock.

internal/app/cache/cache.go
LisaLudique: regarding testing that onEvicted was called: I passed in a test eviction callback function that panics to verify the expected behavior. LMK if it's sufficient - I can remove if it's too hacky, but for now I do not know of another way to test that a function with no return value was called.

regarding shutting down the stream: I had thought we would be using onEvicted as a mechanism for ensuring that anytime an item is removed from the cache, we guarantee that its stream is closed. In my previous PR onEvicted was made configurable so that people could override it with their own behavior, which made it the responsibility of the caller rather than part of the cache implementation. We could shut down the stream when we know an item is removed via TTL but not when the cache is full because LRUCache internally does that, so I think if we want to keep onEvicted as a parameter, it will be on the caller to implement.

internal/app/cache/cache_test.go
LisaLudique: additional context in https://envoyproxy.slack.com/archives/CTULFMN91/p1584655717039600, but I have replaced the call to Eventually with a short time.Sleep to emulate how other cache libraries test expiration.

internal/app/cache/cache.go
LisaLudique: Given that there is a TODO to have the cache implementation guarantee that the stream is opened, I'm inclined to have it also be responsible for closing the stream (and therefore making onEvicted something we set internally and not be configurable - or providing a callback we can default to if the user doesn't provide one).

jessicayuen: > I passed in a test eviction callback function that panics to verify the expected behavior.

I don't think it's hacky, think it's a sensible approach.

> Given that there is a TODO to have the cache implementation guarantee that the stream is opened, I'm inclined to have it also be responsible for closing the stream (and therefore making onEvicted something we set internally and not be configurable - or providing a callback we can default to if the user doesn't provide one).

Yeah, it goes back to the two trains of thoughts we were discussing:
Whether we want
1. orchestrator -> (on evict) -> issue the disconnect with multiplexer,
or
2. cache -> directly issue the disconnect with multiplexer.
cache would still set the `streamOpen` var to false in both scenarios. In scenario 1, it'd be called prior to triggering the callback.

Personally I prefer the first approach as I don't think the cache should be aware of the multiplexer. But that creates some questions of whether the cache should be aware of a `streamOpen` variable all together or if it should be moved to the orchestrator component. I think this was something I could have fleshed out a bit better on the original design. I'm going to think more on this aspect, dropping this here for feedback from both.

internal/app/cache/cache.go
```diff
@@ -1,19 +1,138 @@
...
+		}
+		c.cache.Add(key, resource)
+		return nil, nil
```
jyotimahapatra: should the return be the `resource` just created?

LisaLudique: `resource.requests` should be nil since there was no entry, but I can change that if you think it would be clearer to use `resource.requests`.

internal/app/cache/cache.go
LisaLudique: Like you said I'd rather have eviction happen on reads than writes but in the API cache dump implementation, we could calculate lastModified by taking the expiration time and subtracting the TTL.

internal/app/cache/cache.go
LisaLudique: Makes sense - I've gone ahead and removed all references to streamOpen in the cache implementation.

jessicayuen: thanks, let me know when you push up the change and I'll take another pass.

internal/app/cache/cache.go
jessicayuen: Makes sense.

internal/app/cache/cache.go
LisaLudique: Whoops, I thought I had pushed my changes. It's updated and ready for a look now.

internal/app/cache/cache.go
```diff
@@ -1,19 +1,133 @@
...
+type resource struct {
+	resp           *v2.DiscoveryResponse
+	requests       []*v2.DiscoveryRequest
+	expirationTime time.Time
+}
 
-	// SetResponse sets the cache response and returns the list of open watches.
-	SetResponse(key string, resp envoy_api_v2.DiscoveryResponse) ([]*envoy_api_v2.DiscoveryRequest, error)
+// Callback function for each eviction. Receives the key and cache value when called.
+type onEvictFunc func(key lru.Key, value interface{})
```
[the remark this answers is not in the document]
LisaLudique: I changed the types to string and resource

internal/app/cache/cache.go
```diff
@@ -36,7 +36,7 @@ type resource struct {
 }
 
 // Callback function for each eviction. Receives the key and cache value when called.
-type onEvictFunc func(key lru.Key, value interface{})
+type onEvictFunc func(key string, value resource)
```
jessicayuen: I think you might need to export Resource, otherwise the caller with have no access to it.

PR #31, opened 2020-03-19
RequestNodeFragment result

jyotimahapatra:
This PR implements the result [RequestNodeFragment](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L143)

### DIFF
--- a/internal/app/mapper/mapper.go
+++ b/internal/app/mapper/mapper.go
@@ -204,8 +204,7 @@ func getResult(fragmentRule *rule, node *core.Node) (string, error) {
 		return result, nil
 	}
 
-	stringFragment := fragmentRule.GetResult().GetStringFragment()
-	return stringFragment, nil
+	return fragmentRule.GetResult().GetStringFragment(), nil
 }
 
 func getResultFromRequestNodeFragment(fragmentRule *rule, node *core.Node) (bool, string, error) {
--- a/internal/app/mapper/mapper.go
+++ b/internal/app/mapper/mapper.go
@@ -55,7 +55,10 @@ func (mapper *mapper) GetKey(request v2.DiscoveryRequest) (string, error) {
 				return "", err
 			}
 			if isMatch {
-				result := getResult(fragmentRule)
+				result, err := getResult(fragmentRule, request.GetNode())
+				if err != nil {
+					return "", err
+				}
 				resultFragments = append(resultFragments, result)
 			}
 		}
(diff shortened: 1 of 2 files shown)


jyotimahapatra:
PTAL @jessicayuen @eapolinario @LisaLudique

internal/app/mapper/mapper.go
```diff
@@ -192,9 +195,53 @@ func isNotMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node)
...
+	if found {
+		return result, nil
+	}
+
+	return fragmentRule.GetResult().GetStringFragment(), nil
+}
+
+func getResultFromRequestNodeFragment(fragmentRule *rule, node *core.Node) (bool, string, error) {
+	requestNodeFragment := fragmentRule.GetResult().GetRequestNodeFragment()
+	if requestNodeFragment == nil {
```
jessicayuen: Thoughts on just erroring here, and omitting the found `bool` from the function return?

internal/app/mapper/mapper.go
```diff
@@ -192,9 +195,53 @@ func isNotMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node)
...
+		return false, "", nil
+	}
+
+	nodeField := requestNodeFragment.GetField()
+	var nodeValue = ""
```
jessicayuen: would either do `var nodeValue string` or `nodeValue := ""` (first option would be my preference)

internal/app/mapper/mapper.go
```diff
@@ -192,9 +195,53 @@ func isNotMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node)
...
+		nodeValue = node.GetCluster()
+	case aggregationv1.NodeFieldType_NODE_ID:
+		nodeValue = node.GetId()
+	case aggregationv1.NodeFieldType_NODE_LOCALITY_REGION:
+		nodeValue = node.GetLocality().GetRegion()
+	case aggregationv1.NodeFieldType_NODE_LOCALITY_ZONE:
+		nodeValue = node.GetLocality().GetZone()
+	case aggregationv1.NodeFieldType_NODE_LOCALITY_SUBZONE:
+		nodeValue = node.GetLocality().GetSubZone()
+	}
```
jessicayuen: Can you add a default and error?

internal/app/mapper/mapper.go
```diff
@@ -192,9 +195,53 @@ func isNotMatch(matchPredicate *matchPredicate, typeURL string, node *core.Node)
...
+
+	regexAction := action.GetRegexAction()
+	pattern := regexAction.GetPattern()
+	replace := regexAction.GetReplace()
+
+	reg, err := regexp.Compile(pattern)
+	if err != nil {
+		return false, "", err
+	}
+	return true, reg.ReplaceAllString(nodeValue, replace), nil
```
jessicayuen: If the replace result is empty, I think we should error, `No empty fragments allowed`. I think this will also make it simpler to get rid of the return bool. Assuming you had it to differentiate between empty string case and no result case.

internal/app/mapper/mapper.go
jessicayuen: Nvm, chatted offline. Disregard.

internal/app/mapper/mapper.go
jyotimahapatra: done.

jyotimahapatra:
@jessicayuen I took care of all cases that can produce empty fragments. All such cases now return an error.

internal/app/mapper/mapper.go
eapolinario: shouldn't we assume that the proto will be valid at this point?

internal/app/mapper/mapper.go
eapolinario: does this mean that we should have a tighter validation in https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L123

internal/app/mapper/mapper.go
jyotimahapatra: That is true, but imo its good to have functional completeness of the component.

internal/app/mapper/mapper.go
jyotimahapatra: i think replace string can still be empty and produce a non empty fragment.

internal/app/mapper/mapper.go
jessicayuen: Also for if we ever change the proto and add new type support, but forget to update the mapper

PR #32, opened 2020-03-20
Configuration validator CLI

eapolinario:
This PR adds a standalone binary to help validate configuration files.

Of note, I added a type of test to ensure compilation and execution of the binary itself. I'm calling this an integration test, but I'm not sure if that's the idiosyncratic way of testing stuff like this in go.


Makefile
```diff
@@ -1,23 +1,37 @@
 export SERVICE_NAME=xds-relay
+export GOREPO=${GOPATH}/src/github.com/envoyproxy/${SERVICE_NAME}
+SOURCE_FILES?=$$(go list ./... | grep -v integration)
```
eapolinario: I couldn't find a simpler way of skipping the integration test I'm adding in this PR. Maybe there's a different way of structuring files/packages in such a way that this is not necessary. Feedback on how to do this is more than welcome.

.github/workflows/common.yml
```diff
@@ -35,3 +35,6 @@ jobs:
       - name: Run unit tests
         run: make unit
         working-directory: ${{ env.PACKAGEPATH }}
+      - name: Run configuration validator integration tests
```
jessicayuen: Rather than adding this to the common CI workflow, can you move the cli to a `validator` directory in the [tools](https://github.com/envoyproxy/xds-relay/tree/master/tools) dir, and create a new GitHub checks workflow that only runs this test in the validator directory?

jessicayuen: Or are the tests dependent on other packages in the main app directories?

eapolinario: good idea.

eapolinario: no, the validator integration tests are pretty independent. They build the binary and interact with it directly.

Makefile
```diff
@@ -1,9 +1,14 @@
 export SERVICE_NAME=xds-relay
 
+SOURCE_FILES?=$$(go list ./... | grep -v integration-tests)
```
jyotimahapatra: may be adding a comment about why integration tests need to be added specially will help.

integration-tests/validator-tool_test.go
```diff
@@ -0,0 +1,102 @@
...
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	binaryName = "configuration-validator"
+)
+
+func TestMain(m *testing.M) {
+	// TODO I could not find a more robust way of reaching the root of the repo
```
jyotimahapatra: may be adding an env variable will help find package path?

.github/workflows/common.yml
```diff
@@ -10,7 +10,7 @@ env:
   PACKAGEPATH: ${{ github.workspace }}/go/src/github.com/${{ github.repository }}
   GOLANGCI_LINT_VERSION: v1.23.7
 jobs:
-  test:
+  tests:
```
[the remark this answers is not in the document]
eapolinario: actually, these are arbitrary alphanumeric strings. According to the github actions [documentation](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs), these are known as job ids.

Makefile
eapolinario: will do.

integration-tests/validator-tool_test.go
eapolinario: it's not so much about the package path, but we need to make sure to sure `make` from the root of the repo. At some point during this PR I relied on an env var called `GOREPO` (which relied on `GOPATH`), but since `GOPATH` is [away](https://blog.golang.org/modules2019) I don't think we should use that solution. Do you think it's worth having another env var in our makefile instead? On one hand I'm not fond of the current solution for it can easily break if we decide to move the integration tests to a sub-folder, OTOH, not having another dependency on the environment is better. What are the odds of us moving the config-validator integration tests file?

jyotimahapatra: i might be missing the point, why can't we run the `make compile-validator-tool` in the github workflow?

eapolinario: @jyotimahapatra , that's actually an amazing idea!

eapolinario: on a second thought, we're just shifting the problem from one place to another, i.e., we still need to be at the root of the repo to run the tool in the integration tests.

eapolinario: I mean, we could still run `make compile-validator-tool` in the github workflow, but we still need a way to find the binary location in the tests.

.github/workflows/common.yml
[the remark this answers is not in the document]
jessicayuen: It just changes the UI view in the Checks CI tab.
image

tools/configuration-validator/README.md
```diff
@@ -0,0 +1,22 @@
+## configuration-validator
+
+A tool to help validate aggregation key configuration files
```
jessicayuen:
```suggestion
A tool to help validate xds-relay configuration files.
```

tools/configuration-validator/README.md
```diff
@@ -0,0 +1,22 @@
...
+```
+configuration-validator [flags]
+```
+
+### Options
+
+```
+  -c, --config-file string   path to configuration file
+  -h, --help                 help for configuration-validator
+```
```
jessicayuen: How about the following (removing the `.` before the backticks)?

```
## Synopsis

configuration-validator is a CLI tool used to validate xds-relay configuration
files.

The tool can be embedded in user CI pipelines to ensure that xds-relay is
provisioned correctly.

It currently supports:

* Aggregation Key configurations

## Aggregation Key

The aggregation key yaml configuration is validated against the
[proto file](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto
).  The same proto file is annotated with [protoc-gen-validate constraint
rules](https://github.com/envoyproxy/protoc-gen-validate/#constraint-rules) to
enforce semantic rules. Both the xds-relay initialization and the CLI share the
same validation method.

### Usage

.```
configuration-validator aggregation [flags]
.```

#### Options

.```
  -p, --path string   path to aggregation key configuration file
  -h, --help          help for configuration-validator
.```
```

tools/configuration-validator/main.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/spf13/cobra"
+)
+
+type KeyerConfiguration = aggregationv1.KeyerConfiguration
+
+var (
+	cfgFile string
+
+	validatorCmd = &cobra.Command{
+		Use:   "configuration-validator",
```
jessicayuen: Can you make configuration-validator the root command but add a `aggregation` subcommand? So that we can extend it to support other configurations.

integration-tests/validator-tool_test.go
jyotimahapatra: thinking about this some more. may be use `make compile-validator-tool` in gh workflow and then place the binary in `/usr/local/bin` . In the test reference the path directly instead of `chDir`.

tools/configuration-validator/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+			// Load config file into a valid KeyerConfiguration object and,
+			// in case we encounter an error, print the error string to
+			// stderr.
+			var config KeyerConfiguration
+			err = yamlproto.FromYAMLToKeyerConfiguration(string(yamlFileContent), &config)
+			if err != nil {
+				log.Fatal(err)
+			}
+			os.Exit(0)
```
jyotimahapatra: should doing a `return` here cleanly exit with exit 0?

integration-tests/validator-tool_test.go
eapolinario: I tried this in https://github.com/envoyproxy/xds-relay/pull/32/checks?check_run_id=537737007 and realized that the path `/usr/local/bin` is not available in our github actions workflow. In fact, if we wanted to set it up we'd need to setup all the paths (following something like https://stackoverflow.com/questions/59236611/github-actions-run-golint-on-wihdows/59242962#59242962). But given that we're not investing in relying on GOPATH (since it was deprecated [as of go 1.13](https://blog.golang.org/modules2019)), I don't think we should rely no it.

What do you think?

jyotimahapatra: That was a bummer. Agree that we shouldn't use `GOPATH`. 
I think we have exhausted a lot of approaches. The last one i can think of is still use make from gh actions and place the file in a relative location in the repo based on a env var. pretty gnarly?

eapolinario: at this point I'm not sure we're doing anything better than relying on the current solution (i.e. relying on the relative path).

jyotimahapatra: true. If it makes sense we can put some of this info in the comment. But not a blocker.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,32 @@
+name: integration-tests
```
jessicayuen: Trying to get it to show up as `integration / tests` rather than `integration-tests / integration-tests`. Might want to change the filename to `workflows/integration.yml` too
```suggestion
name: integration
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,32 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+env:
+  PACKAGEPATH: ${{ github.workspace }}/go/src/github.com/${{ github.repository }}
+jobs:
+  integration-tests:
```
jessicayuen:
```suggestion
  tests:
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,32 @@
+name: integration-tests
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
jessicayuen: Should we be adding a path on these to only run in the tools directory? Or can changes in other parts of the app break the tooling?

integration-tests/validator-tool_test.go
```diff
@@ -0,0 +1,94 @@
+package integrationtests
```
jessicayuen: How about `package integration`?

Can you rename the file path to `integration/validator_tool_test.go`? Let's keep with all underscores.

integration-tests/validator-tool_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func TestConfigurationValidatorTool(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "configuration-validator integration tests suite")
+}
+
+var testCases = []TableEntry{
+	{
+		Description: "a positive test",
```
jessicayuen: Since this an integration test, can we try a positive test on a more complex scenario? i.e. over the complete .yaml in the tech spec?

.github/workflows/integration-tests.yml
eapolinario: the file name doesn't really matter, the names that show up in CI are based on the [workflow name](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#name) and the [job name](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idname).

.github/workflows/integration-tests.yml
eapolinario: That's a good idea, but we should also also have the generated pb.go files (i.e., `pkg/api/aggregation/v1/aggregation.pb.go`) there, just to catch the case of someone messing with the proto definitions.

PR #33, opened 2020-03-20
AndResult

jyotimahapatra:
Implements [AndResult](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L137)

### DIFF
--- a/internal/app/mapper/mapper.go
+++ b/internal/app/mapper/mapper.go
@@ -13,7 +13,7 @@ import (
 type matchPredicate = aggregationv1.MatchPredicate
 type rule = aggregationv1.KeyerConfiguration_Fragment_Rule
 type resultPredicate = aggregationv1.ResultPredicate
-type resultPredicateRequestNodeFragment = aggregationv1.ResultPredicate_RequestNodeFragment
+type requestNodeFragment = aggregationv1.ResultPredicate_RequestNodeFragment
 
 // Mapper defines the interface that Maps an incoming request to an aggregation key
 type Mapper interface {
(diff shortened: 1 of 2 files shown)


internal/app/mapper/mapper_test.go
```diff
@@ -423,6 +423,24 @@ var positiveTests = []TableEntry{
...
+		Description: "AnyMatch With result concatenation",
+		Parameters: []interface{}{
+			getAnyMatch(true),
+			getRepeatedResultPredicate1(),
+			clusterTypeURL,
+			nodeid + nodecluster,
+		},
+	},
+	{
+		Description: "yMatch With result concatenation recursive",
```
jessicayuen: `s/yMatch/Match`

internal/app/mapper/mapper.go
```diff
@@ -227,15 +297,25 @@ func getResultFromRequestNodeFragment(fragmentRule *rule, node *core.Node) (bool
...
-		return false, "", fmt.Errorf("RequestNodeFragment Invalid NodeFieldType")
+		return "", fmt.Errorf("RequestNodeFragment Invalid NodeFieldType")
 	}
 
+	return nodeValue, nil
+}
+
+func getResultFragmentFromAction(
+	requestNodeFragment *resultPredicateRequestNodeFragment,
+	node *core.Node) (string, error) {
```
jessicayuen:
```suggestion
func getResultFragmentFromAction(
	requestNodeFragment *resultPredicateRequestNodeFragment,
	node *core.Node,
) (string, error) {
```

jessicayuen: It can probably be all on one line once you shorten the `resultPredicateRequestNodeFragment` name.

internal/app/mapper/mapper_test.go
jyotimahapatra: ah..thanks for spotting

PR #35, opened 2020-03-22
Add logging interface

jessicayuen:
This adds a custom logging interface infront of the zap logger in order
to supply context and friendlier `With` structured logging, ex:

```
    Log.Named("foo-component").With(
        "field1", "value1",
        "field2", "value2",
     ).Error("my error message")
```

Also decouples the 3rd party library code from slipping into other areas
of the application.

### DIFF
--- a/internal/app/server/server.go
+++ b/internal/app/server/server.go
@@ -12,10 +12,12 @@ import (
 	"google.golang.org/grpc"
 )
 
+const defaultLogLevel = "info" // TODO make configurable
+
 // Run instantiates a running gRPC server for accepting incoming xDS-based
 // requests.
 func Run() {
-	logger := log.New().Sugar()
+	logger := log.New(defaultLogLevel)
 
 	// Cursory implementation of go-control-plane's server.
 	// TODO cancel should be invoked by shutdown handlers.
@@ -27,7 +29,7 @@ func Run() {
 	server := grpc.NewServer()
 	listener, err := net.Listen("tcp", ":8080") // #nosec
 	if err != nil {
-		logger.Fatalw("failed to bind server to listener", "err", err)
+		logger.With("err", err).Fatal(ctx, "failed to bind server to listener")
 	}
 
 	api.RegisterEndpointDiscoveryServiceServer(server, gcpServer)
@@ -35,8 +37,8 @@ func Run() {
 	api.RegisterRouteDiscoveryServiceServer(server, gcpServer)
 	api.RegisterListenerDiscoveryServiceServer(server, gcpServer)
 
-	logger.Info("Initializing server at", listener.Addr().String())
+	logger.With("address", listener.Addr()).Info(ctx, "Initializing server")
 	if err := server.Serve(listener); err != nil {
-		logger.Fatalw("failed to initialize server", "err", err)
+		logger.With("err", err).Fatal(ctx, "failed to initialize server")
 	}
 }
--- a/internal/pkg/log/log.go
+++ b/internal/pkg/log/log.go
@@ -2,105 +2,123 @@
 package log
 
 import (
-	"io"
-	"os"
+	"context"
 
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
+	"github.com/envoyproxy/xds-relay/internal/pkg/log/zap"
+
+	z "go.uber.org/zap"
 )
 
-// Options contains all possible log settings.
-type Options struct {
-	// Level configures the log verbosity. Defaults to Debug.
-	Level *zap.AtomicLevel
-	// StacktraceLevel is the level which stacktraces will be emitted. Defaults
-	// to Warn.
-	StacktraceLevel *zap.AtomicLevel
-	// Encoder configures how Zap will encode the output. Defaults to JSON.
-	Encoder zapcore.Encoder
-	// OutputDest controls the destination of the log output. Defaults to
-	// os.Stderr.
-	OutputDest io.Writer
-	// ZapOptions allows passing additional optional zap.Options, ex: Sampling.
-	ZapOptions []zap.Option
+// Logger is the contract for xds-relay's logging implementation.
+//
+// A self-contained usage example looks as follows:
+//
+//    Log.Named("foo-component").With(
+//        "field1", "value1",
+//        "field2", "value2",
+//     ).Error("my error message")
+type Logger interface {
+	// Named adds a sub-scope to the logger.
+	Named(name string) Logger
+
+	// With adds a variadic number of fields to the logging context.
+	// When processing pairs, the first element of the pair is used as the
+	// field key and the second as the field value.
+	//
+	// For example,
+	//
+	//   Log.With(
+	//     "hello", "world",
+	//     "failure", errors.New("oh no"),
+	//     "count", 42,
+	//     "user", User{Name: "alice"},
+	//  ).Info("this is an error message")
+	With(args ...interface{}) Logger
+
+	// Log a message at level Debug, annotated with fields provided through With().
+	Debug(ctx context.Context, msg ...interface{})
+
+	// Log a message at level Info, annotated with fields provided through With().
+	Info(ctx context.Context, msg ...interface{})
+
+	// Log a message at level Warn, annotated with fields provided through With().
+	Warn(ctx context.Context, msg ...interface{})
+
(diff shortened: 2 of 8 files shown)


internal/pkg/log/zap/zap.go
```diff
@@ -0,0 +1,128 @@
...
+		level := zap.NewAtomicLevelAt(zap.WarnLevel)
+		o.StacktraceLevel = &level
+	}
+
+	o.ZapOptions = append(o.ZapOptions, zap.AddStacktrace(o.StacktraceLevel))
+}
+
+// New returns a new zap.Logger configured with the passed Options or their
+// defaults.
+func New(opts ...Opts) *zap.Logger {
```
jyotimahapatra: is the intention here to isolate the zap logger such that it can be replaced by another logger?

jessicayuen: Yeah the zap logger is a 3rd party logger and in general I think it's best practice to isolate it from being called throughout the application code base, especially as our codebase grows and we're using the logger everywhere. Having the interface makes it easier to swap out the underlying 3rd party implementation.

internal/pkg/log/log.go
```diff
@@ -2,105 +2,123 @@
...
+	// Log a message at level Panic, annotated with fields provided through With(), and immediately
+	// panic.
+	Panic(ctx context.Context, msg ...interface{})
+
+	// Log a message at level Fatal, annotated with fields provided through With(), and immediately
+	// call os.Exit.
+	Fatal(ctx context.Context, msg ...interface{})
+
+	// Sync flushes any buffered log entries.
+	Sync() error
```
eapolinario: how is this going to be called?

internal/pkg/log/zap/zap.go
```diff
@@ -0,0 +1,128 @@
...
+	o := &Options{}
+	for _, opt := range opts {
+		opt(o)
+	}
+	o.addDefaults()
+
```
eapolinario: nit: why the empty lines?

internal/pkg/log/log.go
```diff
@@ -2,105 +2,123 @@
...
+
+	if parseLogLevelErr != nil {
+		// Log an invalid log level error and set the default level to info.
+		log.Error("cannot set logger to desired log level")
 	}
+	return &logger{zap: log.Sugar()}
+}
 
-	o.ZapOptions = append(o.ZapOptions, zap.AddStacktrace(o.StacktraceLevel))
+func (l *logger) Named(name string) Logger {
```
jyotimahapatra: are we using a `Named` logger?

jessicayuen: This allows us to isolate components to do something like,
`logger.Named("mapper").Error(ctx, "failed to convert request")
with the output:
`{"level":"info","ts":1585006248.354409,"logger":"mapper","msg":"failed to convert request"}`

internal/pkg/log/log.go
jessicayuen: We'll be calling it in the shutdown handlers of the main gRPC server, https://godoc.org/go.uber.org/zap#hdr-Choosing_a_Logger

internal/app/server/server.go
```diff
@@ -12,10 +12,12 @@ import (
 	"google.golang.org/grpc"
 )
 
+const defaultLogLevel = "info" // TODO make configurable
```
lita: can't this be an env variable?

internal/pkg/log/zap/zap.go
```diff
@@ -0,0 +1,126 @@
+// Package zap sets up a logger using the Zap logging framework.
```
lita: Can we make this comment more explicit? Stating that this package implements the logging interface with the Zap Framework as its implementation?

internal/pkg/log/zap/zap.go
```diff
@@ -0,0 +1,126 @@
...
+	Encoder zapcore.Encoder
+	// OutputDest controls the destination of the log output. Defaults to
+	// os.Stderr.
+	OutputDest io.Writer
+	// ZapOptions allows passing additional optional zap.Options, ex: Sampling.
+	ZapOptions []zap.Option
+}
+
+// Opts allows manipulation of the Zap options.
+type Opts func(*Options)
```
lita: hmm, If you are going to create setters using this type, why does the `Options` struct need to have public attributes?

internal/pkg/log/log.go
```diff
@@ -2,105 +2,123 @@
...
-	}
-	if o.Level == nil {
-		level := zap.NewAtomicLevelAt(zap.DebugLevel)
-		o.Level = &level
-	}
-	if o.StacktraceLevel == nil {
-		level := zap.NewAtomicLevelAt(zap.WarnLevel)
-		o.StacktraceLevel = &level
+// New returns an instance of Logger implemented using the Zap logging framework.
+func New(logLevel string) Logger {
```
lita: I think we should move this to a separate go file and keep the interface standalone with no dependencies to the zap framework. But this is just a suggestion.

internal/pkg/log/log.go
```diff
@@ -2,105 +2,123 @@
...
+	l.WithContext(ctx).zap.Info(args...)
 }
 
-// Encoder configures how the logger will encode the output e.g console, JSON.
-// See Options.Encoder.
-func Encoder(encoder zapcore.Encoder) func(o *Options) {
-	return func(o *Options) {
-		o.Encoder = encoder
-	}
+func (l *logger) Warn(ctx context.Context, args ...interface{}) {
```
lita: I am confused about why this is not implemented in the zap.go file? Maybe I am missing something 😭

internal/pkg/log/log.go
```diff
@@ -2,105 +2,123 @@
 package log
 
 import (
-	"io"
-	"os"
+	"context"
```
lita: This package needs more test coverage. We should have parameterized tests, where we test every logging method.

lita:
The code itself looks good to me. However, the logging framework lacks tests, so it would be good to add more coverage there.

internal/app/server/server.go
jessicayuen: Plan is to have it set via a global config, I want to create a separate PR that'll handle config management soon.

internal/pkg/log/log.go
jessicayuen: Makes sense. Most of the tests right now are in the `zap` package, but I'll add some to cover the interface functions as well.

internal/pkg/log/log.go
jessicayuen: I think there's two sides of the coin to this:
1. I can implement all these in the zap package, but the caller would be calling zap.New(), so even though they'd get the `Logger` interface back...the initial caller would still be aware of the zap package.
2. Implementing it in the log package would make it so the caller isn't aware of the zap package, but then the downside is the interface package is aware of the implementation.

I was going for 2. but I can see it working both ways, I'll go with approach 1 if you think it's cleaner.

internal/pkg/log/zap/zap.go
jessicayuen: Discussed more about this here https://github.com/envoyproxy/xds-relay/pull/35#discussion_r398789804, I'll update the comment.

internal/pkg/log/zap/zap.go
jessicayuen: It doesn't, good catch!

internal/pkg/log/log.go
lita: That is fair. Let's opt for the user experience.

jessicayuen:
@lita Can you take another pass? Most of the tests exist in the [implementation pkg](https://github.com/envoyproxy/xds-relay/blob/4ca8fdd0d492808b1facad53c2e73fc4582d6b41/internal/pkg/log/zap/zap_test.go)

I added a trivial test to make sure that the log in the `log` package initializes successfully under various level circumstances.

It's difficult to test functions like the following since it just calls the underlying zap implementation. Zap doesn't expose it's variables like name, log level, etc. So it's not something I can check against either to make sure it's configured appropriately. I considered overwriting the output file and checking against that but that's not very robust if Zap changes how their log output looks either.
```
func (l *logger) Info(ctx context.Context, args ...interface{}) {
	l.WithContext(ctx).zap.Info(args...)
}
```

PR #38, opened 2020-03-25
ResourceNamesFragment

jyotimahapatra:
Implements [ResourceNamesFragment](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L152-L160)


internal/app/mapper/mapper.go
```diff
@@ -223,15 +230,18 @@ func getResultFromRequestNodeFragmentRule(fragmentRule *rule, node *core.Node) (
...
 
-	return getResultFromRequestNodePredicate(fragmentRule.GetResult(), node)
+	return getResultFromRequestNodePredicate(resultPredicate, node)
 }
 
-func getResultFromAndResultFragmentRule(fragmentRule *rule, node *core.Node) (bool, string, error) {
+func getResultFromAndResultFragmentRule(
+	fragmentRule *rule,
+	node *core.Node,
+	resourceNames []string) (bool, string, error) {
```
eapolinario: I wish `gofmt` could handle formatting of function arguments.

internal/app/mapper/mapper.go
```diff
@@ -282,8 +307,44 @@ func getResultFromAndResultPredicate(resultPredicate *resultPredicate, node *cor
...
+	if resultPredicate == nil {
+		return false, "", nil
+	}
+
+	return getResultFromResourceNamesPredicate(resultPredicate, resoureNames)
+}
+
+func getResultFromResourceNamesPredicate(
+	predicate *resultPredicate,
+	resoureNames []string) (bool, string, error) {
```
eapolinario: typo: `s/resoure/resource/g`

jyotimahapatra: good catch ..thanks

api/protos/aggregation/v1/aggregation.proto
```diff
@@ -151,8 +151,6 @@ message ResultPredicate {
   // [#next-free-field: 4]
   message ResourceNamesFragment {
 
-    NodeFieldType field = 1 [(validate.rules).enum.defined_only = true];
```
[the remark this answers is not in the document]
jyotimahapatra: thanks for spotting this.

internal/app/mapper/mapper.go
jyotimahapatra: is there a different way you have in mind for function structure? I can try that out.

internal/app/mapper/mapper.go
```diff
@@ -271,7 +288,15 @@ func getResultFromAndResultPredicate(resultPredicate *resultPredicate, node *cor
 			resultfragments = resultfragments + fragment
 		}
 
-		found, fragment, err = getResultFromAndResultPredicate(result, node)
+		found, fragment, err = getResultFromResourceNamesPredicate(result, resourceNames)
+		if err != nil {
+			return false, "", err
+		}
+		if found {
+			resultfragments = resultfragments + fragment
```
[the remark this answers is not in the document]
jyotimahapatra: the [proto](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L166-L167) says that the results have to be concatenated.
Is that what you were referring to?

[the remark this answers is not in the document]
eapolinario: no, I was talking about this inefficient way of concatenating strings, as opposed to, for example, using and an array and joining it before returning.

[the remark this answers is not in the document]
jyotimahapatra: makes sense. I did not see that as a problem before, but I agree it could get expensive for repeated concatenations. I'll fix this.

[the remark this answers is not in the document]
jyotimahapatra: done.

go.mod
```diff
@@ -8,8 +8,11 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e
 	github.com/golang/protobuf v1.4.0-rc.4
+	github.com/iancoleman/strcase v0.0.0-20191112232945-16388991a334 // indirect
```
eapolinario: these indirect dependencies are added by the `make compile-protos`. IMO we should just let them go in. What do you think, @jessicayuen?

internal/app/mapper/mapper.go
```diff
@@ -257,33 +274,77 @@ func getResultFromAndResultPredicate(resultPredicate *resultPredicate, node *cor
 	}
 
 	results := resultPredicate.GetAndResult().GetResultPredicates()
-	var resultfragments string
+	var resultfragments []string
```
eapolinario: So, I decided to run a little experiment to benchmark my suggestion and the results were interesting...

I wrote a benchmark test suite that tested 3 methods:
1) naive string concatenation
2) using an array of strings + joining at the end
3) using a [`strings.Builder`](https://golang.org/pkg/strings/#Builder)

The test suite consisted of generating an array of size K containing strings of random lengths between 1 and 50 where `K ∈ {10, 100, 1000, 10K, 100K}` and then concatenating them using each of one of the three methods. Here's the result of one run:
```
❯ go test -bench=.
goos: darwin
goarch: amd64
BenchmarkNaiveConcatenation10-12         2569506               459 ns/op
BenchmarkArray10-12                      1624225               736 ns/op
BenchmarkStringsBuilder10-12             5925367               202 ns/op
BenchmarkNaiveConcatenation100-12          95643             12499 ns/op
BenchmarkArray100-12                      246883              4527 ns/op
BenchmarkStringsBuilder100-12             962344              1251 ns/op
BenchmarkNaiveConcatenation1000-12          1450            747326 ns/op
BenchmarkArray1000-12                      22767             51115 ns/op
BenchmarkStringsBuilder1000-12             82056             14498 ns/op
BenchmarkNaiveConcatenation10K-12             14          72716939 ns/op
BenchmarkArray10K-12                        1014           1042316 ns/op
BenchmarkStringsBuilder10K-12               5199            243196 ns/op
BenchmarkNaiveConcatenation100K-12             1        7027036216 ns/op
BenchmarkArray100K-12                         79          14868540 ns/op
BenchmarkStringsBuilder100K-12               487           2422592 ns/op
PASS
ok      _/tmp   30.048s
```

A few interesting tidbits:
- For arrays of size 10 the naive concatenation is better than using an array
- using a `strings.Builder` is always better than the other 2 methods.
- As the array sizes increase, the gap between array+join vs `strings.Builder` increases reaching almost a 7x difference in the 100K case.

For reference, here's the code for the benchmark:
<details>
<summary>Click to expand</summary>

```
package main

import (
	"math/rand"
	"strings"
	"testing"
)

// Returns an int >= min, < max
func randomInt(min, max int) int {
	return min + rand.Intn(max-min)
}

// Generate a random string of A-Z chars with len = l
func randomString(len int) string {
	bytes := make([]byte, len)
	for i := 0; i < len; i++ {
		bytes[i] = byte(randomInt(65, 90))
	}
	return string(bytes)
}

func benchmarkNaiveConcatenation(k int, b *testing.B) {
	// generate k strings of size up to 20
	xs := make([]string, k)
	for i := 0; i < k; i++ {
		xs = append(xs, randomString(randomInt(1, 20)))
	}
	for n := 0; n < b.N; n++ {
		s := ""
		for _, value := range xs {
			s = s + value
		}
	}
}

func benchmarkStringsBuilder(k int, b *testing.B) {
	// generate k strings of size up to 20
	xs := make([]string, k)
	for i := 0; i < k; i++ {
		xs = append(xs, randomString(randomInt(1, 20)))
	}
	for n := 0; n < b.N; n++ {
		var builder strings.Builder
		for _, value := range xs {
			builder.WriteString(value)
		}
		_ = builder.String()
	}
}

func benchmarkArray(k int, b *testing.B) {
	// generate k strings of size up to 20
	xs := make([]string, k)
	for i := 0; i < k; i++ {
		xs = append(xs, randomString(randomInt(1, 50)))
	}
	for n := 0; n < b.N; n++ {
		var arr []string
		for _, value := range xs {
			arr = append(arr, value)
		}
		_ = strings.Join(arr, "")
	}
}

func BenchmarkNaiveConcatenation10(b *testing.B) {
	benchmarkNaiveConcatenation(10, b)
}

func BenchmarkArray10(b *testing.B) {
	benchmarkArray(10, b)
}

func BenchmarkStringsBuilder10(b *testing.B) {
	benchmarkStringsBuilder(10, b)
}

func BenchmarkNaiveConcatenation100(b *testing.B) {
	benchmarkNaiveConcatenation(100, b)
}

func BenchmarkArray100(b *testing.B) {
	benchmarkArray(100, b)
}

func BenchmarkStringsBuilder100(b *testing.B) {
	benchmarkStringsBuilder(100, b)
}

func BenchmarkNaiveConcatenation1000(b *testing.B) {
	benchmarkNaiveConcatenation(1000, b)
}

func BenchmarkArray1000(b *testing.B) {
	benchmarkArray(1000, b)
}

func BenchmarkStringsBuilder1000(b *testing.B) {
	benchmarkStringsBuilder(1000, b)
}

func BenchmarkNaiveConcatenation10K(b *testing.B) {
	benchmarkNaiveConcatenation(10000, b)
}

func BenchmarkArray10K(b *testing.B) {
	benchmarkArray(10000, b)
}

func BenchmarkStringsBuilder10K(b *testing.B) {
	benchmarkStringsBuilder(10000, b)
}

func BenchmarkNaiveConcatenation100K(b *testing.B) {
	benchmarkNaiveConcatenation(100000, b)
}

func BenchmarkArray100K(b *testing.B) {
	benchmarkArray(100000, b)
}

func BenchmarkStringsBuilder100K(b *testing.B) {
	benchmarkStringsBuilder(100000, b)
}
```
</details>

jyotimahapatra: @eapolinario This is very cool. Due to the experiment i remembered reading something like this [here](https://hermanschaaf.com/efficient-string-concatenation-in-go/) which also put put `bytes.Buffer` in the mix.

Since there are tradeoffs, do you think we should go back to naive concatenation in our usecase?

eapolinario: I added `bytes.Buffer` to the comparison and `strings.Buffer` is still a bit better in all cases. So, in light of this, I think we should switch to `strings.Buffer` (since it's better in all cases, even in the small case).

jyotimahapatra: sounds good. That was a good experiment.

jyotimahapatra: did you mean `strings.Buffer` or `strings.Buffer`?

eapolinario: oops, I meant the latter (there's no `strings.Buffer` in the stdlib).

jyotimahapatra: makes sense..i didn;t see it either. Thanks for confirming

jyotimahapatra:
@eapolinario That was a very nice change. PTAL again.

go.mod
jessicayuen: Yeah, these keep coming back. Makes sense, let's allow them going forward.

PR #45, opened 2020-03-31
Cache the serialized response

jessicayuen:
The cache currently stores the raw discovery response. Store alongside
it the marshaled resource so that we don't have to re-serialize on every
downstream fan out.

### DIFF
--- a/internal/app/cache/cache.go
+++ b/internal/app/cache/cache.go
@@ -7,14 +7,16 @@ import (
 	"sync"
 	"time"
 
+	gcp "github.com/envoyproxy/go-control-plane/pkg/cache"
 	"github.com/golang/groupcache/lru"
+	"github.com/golang/protobuf/ptypes/any"
 
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 )
 
 type Cache interface {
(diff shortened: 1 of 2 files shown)


internal/app/cache/cache.go
```diff
@@ -29,8 +31,14 @@ type cache struct {
 	ttl     time.Duration
 }
 
+// Response stores the raw Discovery Response alongside its marshaled resources.
+type Response struct {
+	raw                v2.DiscoveryResponse
+	marshaledResources []*any.Any
```
lita, line 37: What is the reasoning behind storing both marshaled and raw? Why not just store just the raw response?

jessicayuen: To clarify, do you mean why not just store the marshaled response? (raw being the unserialized response and marshaled being the serialized resources)

Good question though! I don't have a strong opinion on either approach but my main thinking was retaining the raw response allows us to not have opaque logs in other areas of the call stack that may require logging the resources. What you said for debugging purposes! Differentiating between an upstream problem vs a serialization problem, etc.

lita: That makes sense to me. :)

PR #46, opened 2020-04-01
xds client implementation

jyotimahapatra:


internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+		ldsClient:   ldsClient,
+		rdsClient:   rdsClient,
+		edsClient:   edsClient,
+		cdsClient:   cdsClient,
+		callOptions: callOptions,
+	}, nil
+}
+
+// NewMockClient creates a mock implementation for testing
+func NewMockClient(
```
jyotimahapatra, line 101: What are the best practices for exposing Mock functions?

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+	default:
+		return nil, fmt.Errorf("Unsupported request type")
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	signal := make(chan *version, 1)
+	signal <- &version{}
```
lita, line 138: Does xDS protocol require an empty version to start? Not clear why signal needs to buffered but response does not. It would be good to add a comment here.

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+		select {
+		case sig, ok := <-signal:
+			if !ok {
+				return
+			}
+			request.ResponseNonce = sig.nonce
+			request.VersionInfo = sig.version
+			err := doWithTimeout(func() error {
+				return stream.SendMsg(request)
+			}, callOptions.Timeout)
```
lita, line 173: shouldn't the context provide the Deadline here?

internal/app/upstream/client.go
jessicayuen, line 138: Can you explain what we're populating in this channel initially?

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	go send(ctx, &wg, request, response, stream, signal, m.callOptions)
+	go recv(ctx, &wg, response, stream, signal)
+
+	go func() {
+		wg.Wait()
+		closeChannel(signal, response)
```
jessicayuen, line 149: nit: s/closeChannel/closeChannels

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+	ctx context.Context,
+	wg *sync.WaitGroup,
+	request *v2.DiscoveryRequest,
+	response chan *Response,
+	stream grpc.ClientStream,
+	signal chan *version,
+	callOptions CallOptions) {
+	for {
+		select {
+		case sig, ok := <-signal:
```
jessicayuen, line 165: For my understanding, we'll only ever have one signal in the channel right? When the stream is first opened? (because orchestrator only calls `OpenStream` once per aggregated key). Asking because I want to understand the reasoning for the for loop here a bit more.

internal/app/upstream/client_test.go
```diff
@@ -0,0 +1,393 @@
...
+		ctx,
+		createMockLdsClient(errorOnCreate, receiveChan, sendCb),
+		createMockRdsClient(errorOnCreate, receiveChan, sendCb),
+		createMockEdsClient(errorOnCreate, receiveChan, sendCb),
+		createMockCdsClient(errorOnCreate, receiveChan, sendCb),
+		callOptions,
+	)
+}
+
+func createMockLdsClient(
```
lita, line 222: Rather than doing this, I think creating an internal/ or test/utils/fake package with mock xDS clients would be useful unless this is the only module requiring mocked out xDS clients.

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+					wg.Done()
+					return
+				default:
+					response <- &Response{Err: err}
+					wg.Done()
+					return
+				}
+			}
+		case <-ctx.Done():
+			_ = stream.CloseSend()
```
jessicayuen, line 186: Is `_` an error, should we handle it?

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+	for {
+		resp := new(v2.DiscoveryResponse)
+		if err := stream.RecvMsg(resp); err != nil {
+			select {
+			case <-ctx.Done():
+				wg.Done()
+			default:
+				response <- &Response{Err: err}
+			}
+			return
```
jessicayuen, line 208: How come we break out in the error scenario rather than trying to receive future messages?

internal/app/upstream/client.go
lita: There are multiple ways to do it. What you are doing is fine. The other way I have seen is moving this method to a fake.go file that has access to all the private fields/methods.

internal/app/server/server.go
```diff
@@ -33,7 +33,7 @@ func Run() {
 	defer cancel()
 
 	// Initialize upstream client.
-	upstreamClient, err := upstream.NewClient(ctx, upstreamClientURL)
+	upstreamClient, err := upstream.NewClient(ctx, upstreamClientURL, upstream.CallOptions{})
```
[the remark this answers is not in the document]
jyotimahapatra: i wouldn't heap more on the already open PR. I will separately add the configurability.

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
 //
 // The method does not block until the underlying connection is up.
 // Returns immediately and connecting the server happens in background
-// TODO: pass retry/timeout configurations
-func NewClient(ctx context.Context, url string) (Client, error) {
-	return &client{}, nil
+// TODO: pass retry configurations
```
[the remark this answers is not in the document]
jyotimahapatra: yeah. this was supposed to be deleted.

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+func NewClient(ctx context.Context, url string, callOptions CallOptions) (Client, error) {
+	conn, err := grpc.Dial(url, grpc.WithInsecure())
```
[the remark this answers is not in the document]
jyotimahapatra: That is true. However, i'm not sure if that is everything that needs considering. I wouldn't consider the configurability of grpc params in this PR.

internal/app/upstream/client.go
```diff
@@ -55,11 +74,177 @@ type Response struct {
...
+	case listenerTypeURL:
+		stream, err = m.ldsClient.StreamListeners(ctx)
+	case clusterTypeURL:
+		stream, err = m.cdsClient.StreamClusters(ctx)
+	case routeTypeURL:
+		stream, err = m.rdsClient.StreamRoutes(ctx)
+	case endpointTypeURL:
+		stream, err = m.edsClient.StreamEndpoints(ctx)
+	default:
+		return nil, fmt.Errorf("Unsupported request type")
```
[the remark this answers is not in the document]
jyotimahapatra: good point. I will change this to a different error type. Depending on strings probably won;t be strong.

internal/app/upstream/client.go
jyotimahapatra: I intended both to be [unbufferred](https://golang.org/doc/effective_go.html#channels). I'll correct it.

From the [xds protocol](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#ack), the initial request is always empty.

jyotimahapatra: the initial seed is empty version and nonce. I should explicitly initialize them to make the intent clear and add comments.

internal/app/upstream/client.go
jyotimahapatra: From [here](https://github.com/grpc/grpc-go/blob/master/stream.go#L786) looks the err is always nil and non actionable.

internal/app/upstream/client.go
jyotimahapatra: Both [SendMsg](https://github.com/grpc/grpc-go/blob/master/stream.go#L715) and [RecvMsg](https://github.com/grpc/grpc-go/blob/master/stream.go#L734) perform the retries based on the grpc retry params. Since they are already performing the retries, i think we wouldn't benefit from retrying on top of them.

As i have put in `TODO` figuring out grpc retry settings will be a separate PR. When an error happens at this point, the best course would be to open a new stream.

internal/app/upstream/client_test.go
jyotimahapatra: I didn't take the approach since xdsclient mock is used only in one place.

internal/app/upstream/client.go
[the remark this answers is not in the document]
jessicayuen: Sounds good, can you open a issue too?

internal/app/upstream/client.go
jessicayuen: Interesting. Maybe just `stream.CloseSend()` then? (rather than assigning)

jyotimahapatra: that is an artifact of linter:
```
internal/app/upstream/client.go:208:20: Error return value of `stream.CloseSend` is not checked (errcheck)
			stream.CloseSend()
```

internal/app/upstream/client.go
jyotimahapatra: the ctx does not have a deadline at this point. `doWithTimeout ` uses a child context with deadline.

lita: I don't see where it uses the child context? the context doesn't seem to be passed into the doWithTimeout method.

jyotimahapatra: At your time of comment i was using a timer to achieve timeout. I. have since changed the [code](https://github.com/envoyproxy/xds-relay/pull/46/files#diff-c30e80dfa8b61ea3ccdc1ae6d3716890R226) to use ctx and deadline. The github view here is probably stale.

internal/app/upstream/client_test.go
jessicayuen: I think orchestrator could use these rather than reimplementing. These will be valuable to be shareable imo.

jyotimahapatra: I think the mock needed by orchestrator is bit different. It wouldn't need the complex setup provided by this particular mock in discussion. The mock needed by orchestrator would only need a simple `OpenStream` mock for easy testing. WDYT?

jyotimahapatra: I have moved the mock into a reusable package. LMK how does it look.

jyotimahapatra:
@jessicayuen @lita @eapolinario @LisaLudique PTAL. I have incorporated the feedback from the discussion and changed the interface. 
I would like to merge this and then work with the orchestrator and iterate on the interface as needed.

internal/app/upstream/client_test.go
```diff
@@ -0,0 +1,225 @@
...
+		responseChan,
+		func(m interface{}) error { return nil })
+
+	respCh, done, err := client.OpenStream(&v2.DiscoveryRequest{
+		TypeUrl: upstream.ListenerTypeURL,
+		Node:    &core.Node{},
+	})
+	assert.NotNil(t, respCh)
+	assert.Nil(t, err)
+	close(done)
```
jyotimahapatra, line 81: one oddity of the ownership of done channel is that the channel is created in the xdsclient, but expects to be cleaned up in the caller.  We should solve it by changing the ownership of the channel towards the caller, but we can do it in a later iteration once we see value in it.

internal/app/upstream/client.go
```diff
@@ -55,11 +73,184 @@ type Response struct {
...
+	signal chan *version) {
+	for {
+		resp := new(v2.DiscoveryResponse)
+		if err := stream.RecvMsg(resp); err != nil {
+			select {
+			case <-done:
+				cancelFunc()
+				wg.Done()
+			default:
+				close(response)
```
jessicayuen, line 200: Similar here, if this is irrecoverable, should we not just issue wg.Done() too? Do we need the select?

internal/app/upstream/client.go
```diff
@@ -55,11 +73,184 @@ type Response struct {
...
+	// The xds protocol https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#ack
+	// specifies that the first request be empty nonce and empty version.
+	// The origin server will respond with the latest version.
+	signal <- &version{nonce: "", version: ""}
+
+	done := make(chan bool, 1)
+	response := make(chan *v2.DiscoveryResponse, 1)
+
+	var wg sync.WaitGroup
+	wg.Add(2)
```
jessicayuen, line 132: What's the reason that we wait on both recv and send to fail (in the case of an error) or succeed (in the case of a downstream signal to terminate), rather than just either/or? Can this just be `wg.Add(1)`?

internal/app/upstream/client.go
```diff
@@ -25,28 +45,26 @@ type Client interface {
 	// If the timeouts are exhausted, receive fails or a irrecoverable error occurs, the error is sent back to the caller.
 	// It is the caller's responsibility to send a new request from the last known DiscoveryRequest.
 	// Cancellation of the context cleans up all outstanding streams and releases all resources.
-	OpenStream(context.Context, *v2.DiscoveryRequest, string) chan *Response
+	OpenStream(*v2.DiscoveryRequest) (<-chan *v2.DiscoveryResponse, chan bool, error)
 }
 
 type client struct {
```
jessicayuen, line 51: Would also add the logger to this struct and log on the error received upstream since we're no longer propagating downstream.

internal/app/server/server.go
```diff
@@ -38,7 +39,7 @@ func Run(bootstrapConfig *bootstrapv1.Bootstrap,
 	// Initialize upstream client.
 	upstreamPort := strconv.FormatUint(uint64(bootstrapConfig.OriginServer.Address.PortValue), 10)
 	upstreamAddress := net.JoinHostPort(bootstrapConfig.OriginServer.Address.Address, upstreamPort)
-	upstreamClient, err := upstream.NewClient(ctx, upstreamAddress)
+	upstreamClient, err := upstream.NewClient(ctx, upstreamAddress, upstream.CallOptions{Timeout: time.Minute})
```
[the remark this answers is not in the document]
jyotimahapatra: i think i wouldn't address this now since it is unrelated to the core working of the client. I will create an issue and link here.

internal/app/upstream/client.go
```diff
@@ -55,11 +73,184 @@ type Response struct {
...
+	}
+
+	signal := make(chan *version, 1)
+	// The xds protocol https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#ack
+	// specifies that the first request be empty nonce and empty version.
+	// The origin server will respond with the latest version.
+	signal <- &version{nonce: "", version: ""}
+
+	done := make(chan bool, 1)
+	response := make(chan *v2.DiscoveryResponse, 1)
```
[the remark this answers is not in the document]
jyotimahapatra: This is intentional. We are creating a buffered channel since we dont want the goroutines to block on each other. Also, the send and recv calls are lock step and should not happen independently, hence 1. 
Some refs: [ref](https://medium.com/a-journey-with-go/go-buffered-and-unbuffered-channels-29a107c00268), [ref](https://www.ardanlabs.com/blog/2014/02/the-nature-of-channels-in-go.html)

internal/app/upstream/client.go
```diff
@@ -55,11 +73,184 @@ type Response struct {
...
+	signal chan *version,
+	callOptions CallOptions) {
+	for {
+		select {
+		case sig, ok := <-signal:
+			if !ok {
+				return
+			}
+			request.ResponseNonce = sig.nonce
+			request.VersionInfo = sig.version
```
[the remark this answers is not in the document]
jyotimahapatra: makes sense, will remove the pointer. Although fwiw, the pointer is not changing, so the orchestrator map will be unaffected.

internal/app/upstream/client.go
```diff
@@ -55,11 +73,184 @@ type Response struct {
...
+			err := doWithTimeout(ctx, func() error {
+				return stream.SendMsg(request)
+			}, callOptions.Timeout)
+			if err != nil {
+				select {
+				case <-done:
+					cancelFunc()
```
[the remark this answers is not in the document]
jyotimahapatra: There is a nuance here which is coming up because of using closed channel for error signaling: Closing channel on error vs closing channel cleanly.
The `done` channel indicates there was a signal from orchestrator to cleanup. It's not an error condition anymore. The orchestrator has to behave in 2 ways, if its signaling a done, it should not wait on a closed Response channel, because it will be anyway cleanly closed. If there was no signal via done to cleanup and the Response channel was still closed, it indicates an error. 
I was of the opinion that there should be a different signaling mechanism just to indicate error, lmk what you feel about it.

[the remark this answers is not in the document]
jyotimahapatra: I might have misunderstood the question. I changed the implementation some more. LMK if that is aligned with what you are thinking.

internal/app/upstream/client.go
jyotimahapatra: The waitgroup guards the channels which are used in both send and recv goroutines. We need both goroutines to exit before we close the channels. We could block on both channels separately, but both in error and success cases we need both channels to be closed and hence they are together in one group.

jyotimahapatra:
@jessicayuen @lita @eapolinario @LisaLudique Please take another pass. I think its looking much better now.

internal/app/upstream/client.go
```diff
@@ -55,11 +74,179 @@ type Response struct {
...
+func doWithTimeout(ctx context.Context, f func() error, d time.Duration) error {
+	deadlineCtx, cancel := context.WithDeadline(ctx, time.Now().Add(d))
+	errChan := make(chan error, 1)
+	go func() {
+		errChan <- f()
+		close(errChan)
+	}()
+	select {
+	case <-deadlineCtx.Done():
+		cancel()
```
lita, line 242: This cancel context is not passed anywhere..not sure what this does.

jyotimahapatra: i was flummoxed by this error message. Since the cancel operation is idempotent, and `deadlineCtx.Done()` basically means the deadline context is already cancelled, i used the cancel to be a noop and also to get rid of the error.
Screen Shot 2020-04-17 at 9 00 31 AM

lita: I see. If you are calling cancel, I would just `defer cancel()` so you don't have to call it everywhere. `cancel()` is needed to release the resources it is associated with so it cleans itself up.

Also, does it make sense to use context.WithTimeout so it sends a timeout error? If not, then I would rename the private method to be `doWithDeadline`.

internal/app/upstream/client_test.go
```diff
@@ -0,0 +1,221 @@
...
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
+	"github.com/envoyproxy/xds-relay/internal/app/upstream"
+	mock "github.com/envoyproxy/xds-relay/test/mocks/upstream"
+	"github.com/stretchr/testify/assert"
+)
+
+type CallOptions = upstream.CallOptions
+
+func TestOpenStreamShouldReturnErrorForInvalidTypeUrl(t *testing.T) {
```
lita, line 19: I don't see a test here for when the request times out and gets a DeadlineExceeded error

jyotimahapatra: This one is the test: https://github.com/envoyproxy/xds-relay/blob/d3a64bf5d0ec08dd370804a29b820b1a858765e2/internal/app/upstream/client_test.go#L197-L219

We had decided that we don't need to bubble up the error to orchestrator and instead just close the channel.

test/mocks/upstream/client.go
```diff
@@ -0,0 +1,202 @@
+package upstream
```
lita, line 1: Rather than putting this in a giant mock repo, I think it might be better to put this in the same repo as upstream package, like so:

`internal/app/upstream/mock.go`
or
`internal/app/upstream/mock/client.go`

Modules using this mock will still need to import `"github.com/envoyproxy/xds-relay/internal/app/upstream"` anyway to use it. Might as well put them together.

internal/app/upstream/client.go
```diff
@@ -55,11 +74,179 @@ type Response struct {
...
+		resp := new(v2.DiscoveryResponse)
+		if err := stream.RecvMsg(resp); err != nil {
+			select {
+			case <-ctx.Done():
+				// Context was cancelled, hence this is not an erroneous scenario.
+			default:
+				logger.Error(ctx, "Error in RecvMsg %s", err.Error())
+			}
+			cancelFunc()
+			closeChannels(signal, response)
```
lita, line 202: I think it will be cleaner if you break and closeChannels at the end of this method.

internal/app/upstream/client.go
```diff
@@ -55,11 +74,179 @@ type Response struct {
...
+			return
+		}
+		select {
+		case <-ctx.Done():
+			closeChannels(signal, response)
```
lita, line 207: same. break out of the for-loop so that the cleanup method happens in one place.

internal/app/upstream/client.go
lita: I would also call this `shutdown` or `close`, as we might need to add extra steps.

internal/app/upstream/client.go
jyotimahapatra: `context.WithTimeout` makes more sense. Will change `WithDeadline` to `WithTimeout`

internal/app/upstream/client_test.go
lita: But how do you assert that the error was due to a timeout vs if the code changes? Normally, I would test this via a stat and assert that the timeout stat was incremented. 

Can you add a TODO to add that?

internal/app/upstream/client.go
jyotimahapatra: makes sense. will do

internal/app/upstream/client.go
```diff
@@ -55,11 +74,179 @@ type Response struct {
...
+			}
+			request.ResponseNonce = sig.nonce
+			request.VersionInfo = sig.version
+			err := doWithTimeout(ctx, func() error {
+				return stream.SendMsg(request)
+			}, callOptions.Timeout)
+			if err != nil {
+				select {
+				case <-ctx.Done():
+					// Context was cancelled, hence this is not an erroneous scenario.
```
lita, line 169: Just so I understand, this only can happen during a timeout?

test/mocks/upstream/client.go
jyotimahapatra: Sure can do that, i think i moved this to mock package out so that other packages can use based on a previous comment. I like how explicitly other packages can import mock [here](https://github.com/envoyproxy/xds-relay/blob/0ac6a265387b46e81d5f10805ef06e1f8398b82d/internal/app/orchestrator/orchestrator_test.go#L11) [here](https://github.com/envoyproxy/xds-relay/blob/0ac6a265387b46e81d5f10805ef06e1f8398b82d/internal/app/upstream/client_test.go#L13).
LMK if you feel strongly about moving the file into the paths you suggested.

internal/app/upstream/client.go
jyotimahapatra: A timeout from `doWithTimeout` will will cause an error [here](https://github.com/envoyproxy/xds-relay/blob/0ac6a265387b46e81d5f10805ef06e1f8398b82d/internal/app/upstream/client.go#L242).
In the test cases i figured there's racecondition and shutdown could be called when timeout happens. In that case since shutdown causes the parent context cancellation, its prudent to check on it and return. Since the shutdown is already called, there is no point in reporting an error.

internal/app/upstream/client_test.go
jyotimahapatra: Yeah i understand and i previously I meant to expose the error to the caller instead of just closing the channel. Since we have decided to close the channel to report all errors, we have to resort to alternate means to see which error really happens. I will add a TODO to check on a stat once stat is available.

jyotimahapatra: added todo.

internal/app/upstream/client.go
jyotimahapatra: In other words throughout the implementation `ctx.Done()` happens only when `shutdown()` has been called. When goroutines get unblocked and find shutdown has been called, they should exit instead of continuing. I will put the intent in a comment.

jyotimahapatra: added some more comments [here](https://github.com/envoyproxy/xds-relay/blob/f5a366b4816689977fc082da1e07a2f71377cf99/internal/app/upstream/client.go#L48-L50).

test/mocks/upstream/client.go
lita: You can achieve the same result by putting the mocks in the `/mock` folder.

Putting the mock with the struct it is mocking makes it so that the dependencies are clear. If you are not careful, you can also end up with circular dependencies by putting them in one module.

jyotimahapatra: sounds good. I've moved the mocks into `internal/app/upstream/mock`.

internal/app/upstream/client.go
jyotimahapatra: I think this is the remaining one. @lita lmk what you feel about the explanation.

lita: I would add a comment around that the parent context was canceled, probably due to shutdown.

It would also be good to abstract this in a method as these nested selects are hard to read.

```
// Handles the scenario when sendMsg has an error
func handleSendErr(ctx context.Context, err error.Error, cancelFunc func) {
	defer cancelFunc()

	select {
	// If the context is done, then we are probably shutting down. Don't log the error and continue with shutdown.
	case <- ctx.Done():
		return 
	}

	logger.Error(ctx, "Error in SendMsg: %s", err.Error())

}
```

internal/app/upstream/client_test.go
```diff
@@ -0,0 +1,223 @@
...
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
+	"github.com/envoyproxy/xds-relay/internal/app/upstream"
+	mock "github.com/envoyproxy/xds-relay/internal/app/upstream/mock"
+	"github.com/stretchr/testify/assert"
+)
+
+type CallOptions = upstream.CallOptions
+
+func TestOpenStreamShouldReturnErrorForInvalidTypeUrl(t *testing.T) {
```
lita, line 19: You don't have to do this now, but in the future, it might be good to parameterized these tests as you always have to create a MockClient. The only parameters that are changing are CallOptions and the function being passed int.

internal/app/upstream/client.go
lita: Looks like you have the same logic in recv as well which can leverage this code:

https://github.com/envoyproxy/xds-relay/pull/46/files#diff-c30e80dfa8b61ea3ccdc1ae6d3716890R198-R204

lita:
Last set of comments, otherwise looks LGTM. will approve once that select statement is cleaned up.

internal/app/upstream/client_test.go
jyotimahapatra: done. There were 2 buckets of unique types of creation across 6 tests and I pulled them into separate functions.

internal/app/upstream/client.go
jyotimahapatra: pulled the common piece into a function.

internal/app/upstream/client.go
```diff
@@ -55,11 +77,177 @@ type Response struct {
...
+		}
+	}
+	closeChannels(signal, response)
+}
+
+func handleError(ctx context.Context, logger log.Logger, errMsg string, cancelFunc context.CancelFunc, err error) {
+	defer cancelFunc()
+	select {
+	case <-ctx.Done():
+		// Context was cancelled, hence this is not an erroneous scenario.
```
lita, line 210: Shouldn't this be above the case? 🤔 Not sure about go formatting so I could be wrong!

jyotimahapatra: hm...i will look at why lint is not catching it in the next PR.

Issue #41, opened 2020-03-31
Parse env vars and config on server bootstrap


jessicayuen:
Use Cobra or other tooling so that we can start up xds-relay via command line flags or config.
ex:
* `xds-relay -c bootstrap.yml`
* `xds-relay -v INFO --upstream-url example.com`

Initially support log level, upstream url, server addr and port.

jessicayuen:
We'll also want to support supplying aggregation-rules.yml initially.

This is a suggestion:
`xds-relay --aggregation-rules aggregation-rules.yml -c server-config.yml`

jessicayuen:
Ah I'm finding more configuration as I implement the orchestrator, so we'll probably need these for the cache too @LisaLudique! `maxEntries int, onEvicted onEvictFunc, ttl time.Duration`

LisaLudique:
makes sense - I'm wondering if some of these like the cache-related values can live in the bootstrap config itself but not opposed to adding more flags

jessicayuen:
I agree, I would just keep the cache configs in the `-c server-config.yml` configuration. Things like log level and server addr, port can also live in there. More commonly changed config like log level could have a separate overridable flag.

moderation:
This is a fly by comment but interested to see the bootstrap capability. I was surprised to see you using Cobra instead of something a little more modern and light weight like https://pkg.go.dev/github.com/peterbourgon/ff/ffcli?tab=doc.  A recent project Tailscale is using this for its [CLI capability](https://github.com/tailscale/tailscale/search?q=ffcli&unscoped_q=ffcli).  `ffcli` was described by its auther @peterbourgon in this [tweet](https://twitter.com/peterbourgon/status/1166431688705970176) as
> A straightforward and globals-free alternative to spf13/cobra and alecthomas/kingpin.

There was a report on the thread of less code and a reduction of binary size by switching from Cobra to `ffcli`.

jessicayuen:
@moderation This is pretty neat, it wasn't a project on our radar until today. Will definitely check it out, thanks for the mention.

PR #49, opened 2020-04-03
Command line tooling setup and bootstrap config parsing

LisaLudique:
Addresses #41.

This change defines a bootstrap config for the user to pass in preferred settings via a command line tool setup. The values from the config are then used to initialize the various components, e.g. server, cache, orchestrator, logger, upstream client, mapper.


LisaLudique:
@eapolinario @jessicayuen @jyotimahapatra for initial feedback on the bootstrap proto and CLI setup.

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,63 @@
...
+import "validate/validate.proto";
+
+
+// [#next-free-field: 4]
+message Bootstrap {
+    // xds-relay server settings.
+    Server server = 1;
+
+    // Configuration information about the control plane/upstream server.
+    Cluster control_plane_cluster = 2;
```
jessicayuen, line 17: Can we call this `Upstream origin_server` or `Upstream upstream_cluster`? Beyond the MVP, xds-relay could have the capabilities to be a standalone control plane solution and I'm worried this term might get conflated.

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,63 @@
...
+    SocketAddress address = 1;
+}
+
+// [#next-free-field: 3]
+message Logging {
+    // Filepath where logs are emitted. If no filepath is specified, logs will be written to `/dev/null`.
+    string path = 1;
+
+    // The logging level. If no logging level is set, the default is `info`.
+    string level = 2;
```
jessicayuen, line 44: Could we enforce this to be one of `INFO`, `DEBUG`, `WARN`, or `ERROR`?

api/protos/bootstrap/v1/bootstrap.proto
LisaLudique: renamed to origin_server

api/protos/bootstrap/v1/bootstrap.proto
LisaLudique: I limited the values by creating an enum in the proto definition - should we limit the log levels exposed to the user via CLI as well? zap has the following:
```
	// DebugLevel logs are typically voluminous, and are usually disabled in
	// production.
	DebugLevel = zapcore.DebugLevel
	// InfoLevel is the default logging priority.
	InfoLevel = zapcore.InfoLevel
	// WarnLevel logs are more important than Info, but don't need individual
	// human review.
	WarnLevel = zapcore.WarnLevel
	// ErrorLevel logs are high-priority. If an application is running smoothly,
	// it shouldn't generate any error-level logs.
	ErrorLevel = zapcore.ErrorLevel
	// DPanicLevel logs are particularly important errors. In development the
	// logger panics after writing the message.
	DPanicLevel = zapcore.DPanicLevel
	// PanicLevel logs a message, then panics.
	PanicLevel = zapcore.PanicLevel
	// FatalLevel logs a message, then calls os.Exit(1).
	FatalLevel = zapcore.FatalLevel
```

jessicayuen: I think that'd be a good idea, if the error isn't propagated already during parsing of the proto. Our logger supports all the above except `DpanicLevel`

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,69 @@
...
+    Logging logging = 3;
+
+    // Request/response cache settings.
+    Cache cache = 4;
+}
+
+// [#next-free-field: 2]
+message Server {
+    // The TCP address that the xds-relay server will listen on. If not specified, the xds-relay server will not start.
+    SocketAddress address = 1;
```
eapolinario, line 29: ditto.

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// [#next-free-field: 2]
+message Upstream {
+    // The address for the upstream cluster.
+    SocketAddress address = 1;
```
eapolinario, line 35: ditto.

tools/configuration-validator/main.go
```diff
@@ -24,7 +24,7 @@ var (
 	aggregationCmd = &cobra.Command{
 		Use:   "aggregation",
 		Short: "A tool to help validate xds-relay aggregation key configuration files",
-		Long: `aggregation is a comand used to validate aggregation key config files.
+		Long: `aggregation is a command used to validate aggregation key config files.
```
eapolinario, line 27: nice!

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,69 @@
...
+        WARN = 2;
+        ERROR = 3;
+    }
+    Level level = 2;
+}
+
+// [#next-free-field: 3]
+message Cache {
+    // Duration before which a key is evicted from the request/response cache. Zero means no expiration time.
+    google.protobuf.Duration ttl = 1;
```
eapolinario, line 56: add a [Durations](https://github.com/envoyproxy/protoc-gen-validate/#durations) constraint rule to make this required and non-negative.

api/protos/bootstrap/v1/bootstrap.proto
eapolinario: IMO we should catch this error during parsing of the proto.

LisaLudique: The proto will limit the options in the config to INFO/DEBUG/WARN/ERROR, but I meant whether we should prevent the user from entering other log levels via CLI.

integration/bootstrap_test.go
```diff
@@ -0,0 +1,56 @@
...
+			"./integration/testdata/bootstrap_configuration_negative_cache_ttl.yaml",
+			"./integration/testdata/keyer_configuration_complete_tech_spec.yaml",
+			"debug",
+			true,
+			"invalid Cache.Ttl: value must be greater than or equal to 0s",
+		},
+	},
+}
+
+var _ = Describe("Integration tests for bootstrapping xds-relay", func() {
```
eapolinario, line 36: since we won't have positive tests, we can simplify the tests to remove the `wantErr` param. 

I thought a bit about how we could have positive tests, but I couldn't find a simple enough way that doesn't involve waiting or being more chatty in the initialization of the server (which is brittle).

eapolinario:
Can you merge master and tag the integration tests you're adding with the `integration` build tag? (Because of https://github.com/envoyproxy/xds-relay/pull/51)

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,68 @@
...
+
+package bootstrap;
+option go_package = "bootstrap/v1;bootstrapv1";
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/wrappers.proto";
+import "validate/validate.proto";
+
+
+// [#next-free-field: 4]
```
jessicayuen, line 11:
```suggestion
// [#next-free-field: 5]
```

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,68 @@
...
+// [#next-free-field: 3]
+message Cache {
+    // Duration before which a key is evicted from the request/response cache. Zero means no expiration time.
+    google.protobuf.Duration ttl = 1 [(validate.rules).duration = {required: true, gte: {nanos: 0}}];
+
+    // The maximum number of keys allowed in the request/response cache. If unset, no maximum number will be enforced.
+    int32 max_entries = 2;
+}
+
+// [#next-free-field: 4]
```
jessicayuen, line 62:
```suggestion
// [#next-free-field: 3]
```

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,68 @@
...
+message SocketAddress {
+    // The address for this socket. Listeners will bind to the address.
+    string address = 1 [(validate.rules).string = {min_bytes: 1}];
+
+    uint32 port_value = 2 [(validate.rules).uint32 = {lte: 65535}];
+}
```
jessicayuen, line 68: @jyotimahapatra Can you make sure that this is sufficient information to bind upstream?

internal/app/cache/cache.go
```diff
@@ -47,9 +47,6 @@ type Resource struct {
 type OnEvictFunc func(key string, value Resource)
 
 func NewCache(maxEntries int, onEvicted OnEvictFunc, ttl time.Duration) (Cache, error) {
-	if ttl < 0 {
```
jessicayuen, line 50: Did you choose to remove this check because it's checked for in the CLI/proto? Might be still worth having so each component can operate independently.

internal/app/orchestrator/orchestrator.go
```diff
@@ -63,14 +61,21 @@ type orchestrator struct {
...
+	cacheConfig *bootstrapv1.Cache) Orchestrator {
 	orchestrator := &orchestrator{
 		logger:         l.Named(component),
 		mapper:         mapper,
 		upstreamClient: upstreamClient,
 	}
 
-	cache, err := cache.NewCache(cacheMaxEntries, orchestrator.onCacheEvicted, cacheTTL)
+	// Initialize cache.
+	cache, err := cache.NewCache(int(cacheConfig.MaxEntries),
```
jessicayuen, line 76: Rather than performing the int cast, can `MaxEntries` be type int?

internal/app/orchestrator/orchestrator.go
```diff
@@ -63,14 +61,21 @@ type orchestrator struct {
...
+		orchestrator.onCacheEvicted,
+		time.Duration(cacheConfig.Ttl.Nanos)*time.Nanosecond)
```
jessicayuen, line 78: Same with this, any reason we don't just store the ttl as type time.Duration?

api/protos/bootstrap/v1/bootstrap.proto
jyotimahapatra: lgtm

integration/bootstrap_test.go
```diff
@@ -0,0 +1,56 @@
...
+	DescribeTable("table driven integration tests for bootstrapping xds-relay",
+		func(bootstrapConfigFile string, aggregationRulesFile string, logLevel string, wantErr bool, errorMessage string) {
+			dir, err := os.Getwd()
+			Expect(err).To(BeNil())
+
+			// #nosec G204
+			cmd := exec.Command(path.Join(dir, "bin", bootstrapBinaryName),
```
jyotimahapatra, line 43: are we starting up the xds-relay binary ? I think we need a way to bootup and validate the files and not really start all the background connections et al.

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -0,0 +1,68 @@
...
+        WARN = 2;
+        ERROR = 3;
+    }
+    Level level = 2 [(validate.rules).enum.defined_only = true];
+}
+
+// [#next-free-field: 3]
+message Cache {
+    // Duration before which a key is evicted from the request/response cache. Zero means no expiration time.
+    google.protobuf.Duration ttl = 1 [(validate.rules).duration = {required: true, gte: {nanos: 0}}];
```
jyotimahapatra, line 56: im not sure we should expose `google.protobuf.Duration` in config.
What are your thoughts around putting in a way for folks to specify 1s, 1h, 1ns ,kind of format?

jyotimahapatra: nvm, saw the test file.

internal/app/orchestrator/orchestrator_test.go
```diff
@@ -27,6 +29,13 @@ func TestNew(t *testing.T) {
 	}
 	requestMapper := mapper.NewMapper(&config)
 
-	orchestrator := New(context.Background(), log.New("info"), requestMapper, upstreamClient)
+	cacheConfig := bootstrapv1.Cache{
+		Ttl: &duration.Duration{
```
jyotimahapatra, line 33: shouldnt this be `time.duration`?

internal/app/cache/cache.go
LisaLudique: Can you clarify why the check should remain?
I think if we're going to be opinionated about the proto validation being responsible for enforcing potential values we should be consistent and remove checks from other places in the code.

internal/app/orchestrator/orchestrator.go
LisaLudique: https://developers.google.com/protocol-buffers/docs/proto#scalar int isn't an available type

internal/app/orchestrator/orchestrator.go
LisaLudique: I don't think it's possible to set the proto type to time.Duration. We could leverage gogo to convert more seamlessly between the two: https://github.com/gogo/protobuf/blob/master/extensions.md#more-canonical-go-structures.

internal/app/orchestrator/orchestrator_test.go
LisaLudique: See https://github.com/envoyproxy/xds-relay/pull/49#discussion_r406479958.

internal/app/cache/cache.go
jessicayuen: My thinking around this was that there's no direct relationship between the proto and this constructor. i.e. Nothing preventing us from instantiating a cache outside of the orchestrator and giving it a TTL < 0 in the future, and have the error go uncaught. But since this is an internal package, I could see the argument for it not being necessary since a mistake like that isn't likely to happen. No strong opinions about this, I'll defer to you to make the final call!

internal/app/orchestrator/orchestrator.go
jessicayuen: Got it.

integration/bootstrap_test.go
LisaLudique: added 'validate' and 'serve' modes. Right now they do the exact same things, except 'validate' doesn't call .Serve().

internal/app/cache/cache.go
LisaLudique: added it back

PR #50, opened 2020-04-07
More precise locking for cache.Fetch

LisaLudique:
This PR is an attempt to make sharing of the cache resources more efficient amongst different go-routines.

### DIFF
--- a/internal/app/cache/cache.go
+++ b/internal/app/cache/cache.go
@@ -73,9 +73,9 @@ func NewCache(maxEntries int, onEvicted onEvictFunc, ttl time.Duration) (Cache, 
 }
 
 func (c *cache) Fetch(key string) (*Response, error) {
-	c.cacheMu.Lock()
-	defer c.cacheMu.Unlock()
+	c.cacheMu.RLock()
 	value, found := c.cache.Get(key)
+	c.cacheMu.RUnlock()
 	if !found {
 		return nil, fmt.Errorf("no value found for key: %s", key)
 	}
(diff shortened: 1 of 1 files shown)


internal/app/cache/cache.go
```diff
@@ -85,8 +85,23 @@ func (c *cache) Fetch(key string) (*Response, error) {
...
+		value, found = c.cache.Get(key)
+		if !found {
+			// The entry was already evicted.
+			return nil, nil
+		}
+		resource, ok = value.(Resource)
+		if !ok {
+			return nil, fmt.Errorf("unable to cast cache value to type resource for key: %s", key)
+		}
+		if resource.isExpired(time.Now()) {
```
jessicayuen, line 99: We're checking `isExpired` again in the case of a recent concurrent `set` right?

jessicayuen: Would just document this a bit, but approach makes sense to me.

internal/app/cache/cache.go
```diff
@@ -85,8 +85,23 @@ func (c *cache) Fetch(key string) (*Response, error) {
...
+			c.cache.Remove(key)
+			return nil, nil
+		}
+		// The entry should no longer be evicted.
+		return resource.resp, nil
```
jessicayuen, line 104: Prob can delete this due to redundancy with the next line outside of the loop.

internal/app/cache/cache.go
LisaLudique: yep, added some comments

PR #51, opened 2020-04-07
Use tags to mark integration tests

eapolinario:
This PR removes the hack to list test files using `go list` in favor of the more idiomatic way of setting [build tags](https://golang.org/pkg/go/build/#hdr-Build_Constraints) to segment integration tests.

Build tags are opt-in, that's why they are useful for stuff like integration tests. This means that in order for tagged files to show up we have to specify the tags, so in this case we tag all integration tests as `integration` and specify that tag in the `integration-tests` make rule.

### DIFF
--- a/integration/validator_tool_test.go
+++ b/integration/validator_tool_test.go
@@ -1,3 +1,5 @@
+// +build integration
+
 package integration
 
 import (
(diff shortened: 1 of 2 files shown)


integration/validator_tool_test.go
```diff
@@ -1,3 +1,5 @@
+// +build integration
```
jyotimahapatra, line 1: awesome.

PR #52, opened 2020-04-08
Return Resource on cache Fetch

jessicayuen:
* Return `Resource` rather than `Response` on fetch. Orchestrator needs
  the Resource object afterall in order to identify and notify watches.
* Export `Resource` and `Response` members that will be used by calling
  classes.
* Change the marshaled resource type to `[][]byte`, which is what
  go-control-plane expects when constructing the Response.
* Remove the comment about notify watches, responsibility has moved to
  Orchestrator.

### DIFF
--- a/internal/app/cache/cache.go
+++ b/internal/app/cache/cache.go
@@ -7,16 +7,16 @@ import (
 	"sync"
 	"time"
 
+	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
+	gcp_types "github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	gcp "github.com/envoyproxy/go-control-plane/pkg/cache/v2"
 	"github.com/golang/groupcache/lru"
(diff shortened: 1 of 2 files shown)


internal/app/cache/cache.go
```diff
@@ -33,14 +33,14 @@ type cache struct {
...
+	MarshaledResources []gcp_types.MarshaledResource
 }
 
 type Resource struct {
-	resp           *Response
-	requests       []*v2.DiscoveryRequest
-	expirationTime time.Time
+	Resp           *Response
+	Requests       []*v2.DiscoveryRequest
+	ExpirationTime time.Time
```
LisaLudique, line 43: Does ExpirationTime need to be exported?

jessicayuen: Not for the orchestrator, but I'm sure that we'll need it for our admin APIs later when we dump what's in the cache.

LisaLudique: makes sense!

PR #53, opened 2020-04-08
Orchestrator implementation

jessicayuen:
Add the end-to-end request management workflow.
See comments inline for details.

Three open issues to follow up on (also inline in code):
https://github.com/envoyproxy/xds-relay/issues/71
https://github.com/envoyproxy/xds-relay/issues/70
https://github.com/envoyproxy/xds-relay/issues/68


internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+		//
+		// Locking is necessary here so that a simultaneous downstream request
+		// that maps to the same aggregated key doesn't result in two upstream
+		// streams.
+		o.upstreamResponseChannelseMu.Lock()
+		if o.upstreamResponseChannels[aggregatedKey] == nil {
+			upstreamResponseChan := o.upstreamClient.OpenStream(ctx, &req)
+			// Spin up a go routine to watch for upstream responses.
+			// One routine is opened per aggregate key.
+			go o.watchUpstream(ctx, aggregatedKey, upstreamResponseChan)
```
jessicayuen, line 173: TODO: During cache expiry, will need to clean up these routines.

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
-func (c *orchestrator) onCacheEvicted(key string, resource cache.Resource) {
-	// TODO implement.
+// watchResponse is intended to be called in a goroutine, to receive incoming
+// responses and fan out to downstream clients.
+func (o *orchestrator) watchUpstream(
+	ctx context.Context,
+	aggregatedKey string,
+	responseChannel <-chan *upstream.Response,
+) {
+	for {
```
lita, line 194: This needs a shutdown handler. One option is to listen for a shutdown channel and break out of the for loop.

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+			time.Sleep(1 * time.Second)
+		}
+	}
+}
+
+// fanout pushes the response to the response channels of all open downstream
+// watches.
+func (o *orchestrator) fanout(resp *cache.Response, watchers []*gcp.Request) {
+	for _, watch := range watchers {
+		channel := o.downstreamResponseChannels[watch]
```
lita, line 249: Needs a lock for read.

internal/app/upstream/client.go
```diff
@@ -25,7 +25,7 @@ type Client interface {
 	// If the timeouts are exhausted, receive fails or a irrecoverable error occurs, the error is sent back to the caller.
 	// It is the caller's responsibility to send a new request from the last known DiscoveryRequest.
 	// Cancellation of the context cleans up all outstanding streams and releases all resources.
-	OpenStream(context.Context, *v2.DiscoveryRequest, string) chan *Response
+	OpenStream(context.Context, *v2.DiscoveryRequest) <-chan *Response
```
lita, line 28: Just so I understand, OpenStream is the one adding Responses to the receiver only channel? Thus the upstream/client.go has ownership of the `chan *Response` ?

jessicayuen: Yep, correct. upstream/client pushes responses to the receiver-only channel and has ownership of it.

@jyotimahapatra has a [WIP implementation](https://github.com/envoyproxy/xds-relay/pull/46) of it. Jyoti, I changed this to a receiver only channel so that orchestrator can read from it, let me know your thoughts as well.

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+
+	aggregatedKey, err := o.mapper.GetKey(req)
+	if err != nil {
+		// Can't map the request to an aggregated key. Log and continue to
+		// propagate the response upstream without aggregation.
+		o.logger.With("err", err).With("req", req).Error(ctx, "failed to get aggregated key")
+		// Mimic the aggregated key.
+		// TODO (insert issue) This key needs to be made more granular to
+		//      uniquely identify a request. User-specified defaults?
+		aggregatedKey = fmt.Sprintf("%s%s_%s", unaggregatedPrefix, req.GetNode().GetId(), req.GetTypeUrl())
```
jyotimahapatra, line 127: using `req.GetNode().GetId()` will cause many such keys. Thoughts?

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+		// If we fail to register the watch, we need to kill this stream by
+		// closing the response channel.
+		o.logger.With("err", err).With("key", aggregatedKey).With(
+			"req", req).Error(ctx, "failed to add watch")
+		close(o.downstreamResponseChannels[&req])
+		// Clean up.
+		o.downstreamResponseChannelsMu.Lock()
+		delete(o.downstreamResponseChannels, &req)
+		o.downstreamResponseChannelsMu.Unlock()
+		return nil, nil
```
jyotimahapatra, line 142: we should probably return a closed channel?

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+			upstreamResponseChan := o.upstreamClient.OpenStream(ctx, &req)
+			// Spin up a go routine to watch for upstream responses.
+			// One routine is opened per aggregate key.
+			go o.watchUpstream(ctx, aggregatedKey, upstreamResponseChan)
+			o.upstreamResponseChannels[aggregatedKey] = upstreamResponseChan
+		}
+		o.upstreamResponseChannelseMu.Unlock()
+	}
+
+	return o.downstreamResponseChannels[&req], nil
```
jyotimahapatra, line 179: would we not need the cleanup?

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+		delete(o.downstreamResponseChannels, &req)
+		o.downstreamResponseChannelsMu.Unlock()
+		return nil, nil
+	}
+
+	// Check if we have a cached response first.
+	cached, err := o.cache.Fetch(aggregatedKey)
+	if err != nil {
+		// Log, and continue to propagate the response upstream.
+		o.logger.With("err", err).With("key", aggregatedKey).Error(ctx, "failed to fetch aggregated key")
```
jyotimahapatra, line 149: Should the log type be Warn instead?

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+	}
+
+	if cached != nil && cached.Resp != nil {
+		// If we have a cached response, immediately return the result.
+		o.downstreamResponseChannels[&req] <- gcp.Response{
```
jyotimahapatra, line 154: im thinking about this scenario:
new sidecar -> value available in cache -> send immediate response

response gets applied on sidecar -> sidecar asks for another update 

At this point, we have to send back a channel which has nothing, so that it can be fulfilled later. 
Can you describe briefly the code path that enables the scenario?

internal/app/upstream/client.go
jyotimahapatra: yeah. this is the intended behavior.

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
 //
 // Cancel is an optional function to release resources in the producer. If
 // provided, the consumer may call this function multiple times.
-func (c *orchestrator) CreateWatch(req gcp.Request) (chan gcp.Response, func()) {
-	// TODO implement.
-	return nil, nil
+func (o *orchestrator) CreateWatch(req gcp.Request) (chan gcp.Response, func()) {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
lita: The interface of gcp does not have the context as a parameter. https://github.com/envoyproxy/go-control-plane/blob/f070f3d1d73b03f00369ee424d38077944e7b984/pkg/cache/v2/cache.go#L43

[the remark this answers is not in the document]
jyotimahapatra: Yeah ctx can be passed from constructor. We will need that to make sure clean up happens on shutdown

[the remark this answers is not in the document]
lita: How would you pass the ctx from the constructor? The constructor doesn't call CreateWatch.

Apologies if I am misinterpreting you. But the only way I can see getting the context from the constructor is to store the context in the orchestrator object, but that is bad practice. [The context should be in the scope of the request, not the creation of the constructor.](https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation) In an ideal world, CreateWatch should pass a context as a parameter. 

I agree that a cancel function needs to be set up and have each of the watch listen for it. Unclear what the best mechanism is as go-control-plane doesn't pass around ctx as a parameter in their interfaces.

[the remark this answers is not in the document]
jyotimahapatra: Thanks for the link. I will explore the possibility of using the `tomb` package. However,fortunately, because the xdsclient is initialized in [server.go](https://github.com/envoyproxy/xds-relay/blob/master/internal/app/server/server.go#L36), i think we are covered here. Nonetheless the same issue will be a problem in xds client.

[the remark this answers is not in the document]
jyotimahapatra: Looks like g-c-p [saves the context](https://github.com/envoyproxy/go-control-plane/blob/master/pkg/server/v2/server.go#L87).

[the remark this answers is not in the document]
jyotimahapatra: Thinking about this more, the [OpenStream](https://github.com/envoyproxy/xds-relay/blob/master/internal/app/upstream/client.go#L28) takes a context, which needs to be wired up to global context, so that cancellation caused the waiting goroutines for [RecvMsg](https://github.com/grpc/grpc-go/blob/b02de00073bfccaef63816bfd3ac296e9abc371f/stream.go#L126-L132) to exit.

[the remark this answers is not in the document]
lita: This isn't ideal. If you look at the [context package](https://golang.org/pkg/context/), they also state `Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx`. 

I think it is worth refactoring go-control-plane such that we don't store it on the server but either pass it in or store a `Cancel` method in the server which we can call. But that is up to you.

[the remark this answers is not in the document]
jyotimahapatra: Yeah I agree that context shouldn't be saved. I'm suggesting that we need some way to cleanup resources eg connections ,goroutines etc during shutdown and other scenarios.

internal/app/orchestrator/orchestrator.go
jessicayuen: done, https://github.com/envoyproxy/xds-relay/pull/53/files#diff-2958638164db92742469d2bb08d0001aR232

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +105,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+		// downstream client, initialize a channel to feed future responses.
+		o.downstreamResponseChannels[&req] = make(chan gcp.Response)
+	}
+	o.downstreamResponseChannelsMu.Unlock()
+
+	aggregatedKey, err := o.mapper.GetKey(req)
+	if err != nil {
+		// Can't map the request to an aggregated key. Log and continue to
+		// propagate the response upstream without aggregation.
+		o.logger.With("err", err).With("req", req).Error(ctx, "failed to get aggregated key")
```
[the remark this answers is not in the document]
jessicayuen: I changed this to log the type and node object for now.

internal/app/orchestrator/orchestrator.go
jessicayuen: I'd like to open an issue and work on this unconventional path, "non-aggregated" path in a separate PR.  I think it needs some more discussion.

internal/app/orchestrator/orchestrator.go
jessicayuen: Added.

internal/app/orchestrator/orchestrator.go
jessicayuen: Please take a look again and see if the workflow makes more sense.

new sidecar -> value in cache -> immediate response -> watch is still opened if it doesn't already exist

new response from upstream -> fan out to watches -> channel gets populated.

internal/app/orchestrator/orchestrator.go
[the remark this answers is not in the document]
jessicayuen: I've been thinking more about this and have a [TODO](https://github.com/envoyproxy/xds-relay/pull/53/files#diff-a3a788e64cea038d60dbf226fa6e275eR37) in the upstream map clean up portion that relates to this.

An approach I've been considering is having `OpenStream` from the upstream client supply not only the receiver channel, but a "done" channel. 

```
OpenStream(context.Context, *v2.DiscoveryRequest, string) <-chan *Response, chan bool
```

when shutdown occurs, orchestrator will close the bool channel. Orchestrator [already has to do this](https://github.com/envoyproxy/xds-relay/pull/53/files#diff-a3a788e64cea038d60dbf226fa6e275eR30) to clean up the go routines during [watchUpstream](https://github.com/envoyproxy/xds-relay/pull/53/files#diff-2958638164db92742469d2bb08d0001aR232). Rather than initializing a new channel, we can just assign it the one returned from `OpenStream`.

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +110,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+}
+
+// fanout pushes the response to the response channels of all open downstream
+// watches.
+func (o *orchestrator) fanout(resp *cache.Response, watchers []*gcp.Request) {
+	for _, watch := range watchers {
+		if channel, ok := o.downstreamResponseMap.get(watch); ok {
+			channel <- convertToGcpResponse(resp, *watch)
+		}
+	}
```
gnz00, line 249: Would it be worth sending these responses in parallel?

internal/app/orchestrator/orchestrator.go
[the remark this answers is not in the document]
jyotimahapatra: I could be wrong, but one aspect to consider while using channels for cancellation is that channels can have only one receiver. If there are more goroutines than one, we have to think about how to cancel all of them using the same channel. A context on the other hand can be used to send cancellation signal to multiple goroutines.

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +110,172 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+	// Check if we have a cached response first.
+	cached, err := o.cache.Fetch(aggregatedKey)
+	if err != nil {
+		// Log, and continue to propagate the response upstream.
+		o.logger.With("err", err).With("key", aggregatedKey).Warn(ctx, "failed to fetch aggregated key")
+	}
+
+	if cached != nil && cached.Resp != nil {
+		// If we have a cached response, immediately push the result to the
+		// response channel.
```
jyotimahapatra, line 151: If we have a cached response and the version is equal to `discoveryRequest.VersionInfo` we shouldn;t send a response back. Instead just send back an open channel.

internal/app/orchestrator/orchestrator.go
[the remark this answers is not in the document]
jessicayuen: Made a simple test here that shows closing the channel can notify multiple receivers: https://play.golang.org/p/QHhcOZXNLSa

internal/app/orchestrator/orchestrator.go
jessicayuen: Added https://github.com/envoyproxy/xds-relay/issues/56 and linked.

internal/app/orchestrator/orchestrator.go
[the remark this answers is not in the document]
jyotimahapatra: Thanks for the experiment. It could work out then.
I was using https://play.golang.org/p/NQGO5-jCVz to make my observations. Observe how only one goroutine gets all the values.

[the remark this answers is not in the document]
jessicayuen: Yeah, `close()` is a special case that seems to work differently than values being pushed to the channel (which only the faster receiver can read from)

internal/app/orchestrator/orchestrator.go
jessicayuen: Done, thanks for the suggestion!

internal/app/orchestrator/orchestrator.go
jessicayuen: Yeah, good catch. I'll make the change and add a test case.

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +110,176 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+
+	// Check if we have a upstream stream open for this aggregated key. If not,
+	// open a stream with the representative request.
+	//
+	// Locking is necessary here so that a simultaneous downstream request
+	// that maps to the same aggregated key doesn't result in two upstream
+	// streams.
+	o.upstreamResponseMap.mu.Lock()
+	if _, ok := o.upstreamResponseMap.responseChannel[aggregatedKey]; !ok {
+		upstreamResponseChan := o.upstreamClient.OpenStream(ctx, &req)
```
jyotimahapatra, line 163: i'm still a bit unclear about how to split the responsibilities between `ctx` and `done` channel.
The grpc streams are based on the `ctx` and `RecvMsg` will return when ctx is cancelled. We still need a way to cancel the context for `RecvMsg` to stop blocking and clean up the goroutine. Is there a different way you envision the stream can be cancelled?

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +110,176 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+	done <-chan bool,
+) {
+	for {
+		select {
+		case x := <-responseChannel:
+			if x.Err != nil {
+				// A problem occurred fetching the response upstream, log and
+				// return the most recent cached response, so that the
+				// downstream will reissue the discovery request.
+				o.logger.With("err", x.Err).With("key", aggregatedKey).Error(ctx, "upstream error")
```
jyotimahapatra, line 195: should we still progress if an `Err` was received?

internal/app/orchestrator/orchestrator.go
```diff
@@ -90,16 +110,176 @@ func New(ctx context.Context, l log.Logger, mapper mapper.Mapper, upstreamClient
...
+					// Goldenpath.
+					o.fanout(cached.Resp, cached.Requests)
+				}
+			}
+		case <-done:
+			// Exit when signaled that the stream has closed.
+			return
+		default:
+			// Save some processing power.
+			time.Sleep(1 * time.Second)
```
jyotimahapatra, line 237: iiuc, we don't need this. The channels will be blocked anyway.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: also i think `OpenStream` should be called again to reinitiate the stream.

internal/app/cache/cache.go
```diff
@@ -110,7 +110,7 @@ func (c *cache) Fetch(key string) (*Resource, error) {
 func (c *cache) SetResponse(key string, resp v2.DiscoveryResponse) ([]*v2.DiscoveryRequest, error) {
 	c.cacheMu.Lock()
 	defer c.cacheMu.Unlock()
-	marshaledResources, err := marshalResources(resp.Resources)
+	marshaledResources, err := MarshalResources(resp.Resources)
```
lita, line 113: Wondering why this needs to be public.

internal/app/orchestrator/orchestrator.go
```diff
@@ -50,6 +54,13 @@ const (
 // more details.
 type Orchestrator interface {
 	gcp.Cache
+
+	// shutdown takes an aggregated key and shuts down the go routine watching
+	// for upstream responses.
+	//
+	// This is currently used by tests to clean up channels, but can also be
+	// used by the main shutdown handler.
+	shutdown(string)
```
lita, line 63: Can we add a name for the parameter? I am guessing the string value is the key?

internal/app/orchestrator/orchestrator.go
jessicayuen: I'm still operating under the model than an error is not a non-recoverable state for the stream. I think if it's non-recoverable, closing the channel makes more sense. Let's discuss more in our meeting tomorrow.

internal/app/orchestrator/orchestrator.go
jessicayuen: Sounds good, this will make the tests faster too.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: A stream could be overloaded or in a bad state, but opening a new stream in that scenario would be ok, no?

internal/app/cache/cache.go
jessicayuen: This is used by the orchestrator tests in order to compare the marshaled response with the mock desired one: https://github.com/envoyproxy/xds-relay/pull/53/files#diff-2c2ed0f611d2764f6a076c08ee2b18f7R88. Open to suggestions on how to better do this!

internal/app/orchestrator/orchestrator.go
lita: That is not necessarily true. You can have OpenStream take in a shutdown channel and then have it not listen to RecvMsg anymore and exit the for-loop. Ideally, we can push an EOF to envoy or a GOAWAY header so that envoy will close the connection before exiting the loop, but I don't think that is as important.

jyotimahapatra: based on [this](https://github.com/grpc/grpc-go/blob/master/stream.go#L114-L117) RecvMsg is a blocking call and will block irrespective. the only way to stop that is by [cancelling](https://github.com/grpc/grpc-go/blob/master/stream.go#L132) the context

lita: You can leverage `select` to listen for both RecvMsg or the cancel channel.

jyotimahapatra: I think there's some more details that are important here. The ownership of the `done` channel is with upstream. This means upstream can communicate to the orchestrator to shutdown. However i don't the ownership model be inverted like that. When orchestrator figures that a key is not important anymore or ttl expires and the upstream stream has to be released, or in any other clean up case, how does orchestrator signal that to the upstream client? I'm not necessarily suggesting to use ctx, but we need something for the usecase i mentioned.

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +115,173 @@ func New(ctx context.Context,
...
+						o.logger.With("key", aggregatedKey).
+							Error(ctx, "attempted to fan out with no cached response")
+					}
+				} else {
+					// Goldenpath.
+					o.fanout(cached.Resp, cached.Requests)
+				}
+			}
+		case <-done:
+			// Exit when signaled that the stream has closed.
```
jyotimahapatra, line 238: A stream should never be closed. The reason a stream was open was because a sidecar asked for a resource. If the key cannot be found on upstream, we should just retry(with backoff) .

internal/app/orchestrator/orchestrator.go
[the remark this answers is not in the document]
jyotimahapatra: Reading about context cancellation, there's so many contradicting views. [here](https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation), [here](https://twitter.com/davecheney/status/894717817886433280), [here](https://faiface.github.io/post/context-should-go-away-go2/).

internal/app/cache/cache.go
lita: No worries. This makes sense. 👍

internal/app/orchestrator/orchestrator.go
lita: I think you can have two channels for this responsibility. One for the upstream to communicate that it is shutting down and the orchestrator to communicate up that the stream is no longer valid.  I am fine with you using ctx for this as well, but I am not sure how you will be able to do it without storing ctx on the datastructure. Ideally, it would be great if CreateWatch accepted a context.

internal/app/orchestrator/orchestrator.go
jessicayuen: I'm using this during shutdown and testing, please see the updated commits.

internal/app/orchestrator/orchestrator.go
jessicayuen: Changed this to take a context and shutdown all upstream streams.

internal/app/orchestrator/orchestrator.go
jessicayuen: I believe we've agreed on a solution offline and this is reflected in the upstream client implementation with the returned callback, but let me know if that's not the case.

internal/app/orchestrator/orchestrator_test.go
```diff
@@ -2,19 +2,99 @@ package orchestrator
...
+	assert.NoError(t, err)
+
+	var config aggregationv1.KeyerConfiguration
+	err = yamlproto.FromYAMLToKeyerConfiguration(string(bytes), &config)
+	assert.NoError(t, err)
+
+	return mapper.NewMapper(&config)
+}
+
+func assertEqualResources(t *testing.T, got gcp.Response, expected v2.DiscoveryResponse, req gcp.Request) {
```
lita, line 89: In the future, you might want to share this and add it as a global testing utils function.

lita:
Since the client is merged, do you plan on adding an integration test in a different PR?

jessicayuen:
Yeah, going to create an issue to work on integration or E2E tests.

internal/app/orchestrator/orchestrator_test.go
jessicayuen: Makes sense. We should refactor this out when/if there are multiple call sites.

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,183 @@ func New(ctx context.Context,
...
+	cached, err := o.cache.Fetch(aggregatedKey)
+	if err != nil {
+		// Log, and continue to propagate the response upstream.
+		o.logger.With("err", err).With("key", aggregatedKey).Warn(ctx, "failed to fetch aggregated key")
+	}
+
+	if cached != nil && cached.Resp != nil && cached.Resp.Raw.GetVersionInfo() != req.GetVersionInfo() {
+		// If we have a cached response and the version is different,
+		// immediately push the result to the response channel.
+		go func() { responseChannel <- convertToGcpResponse(cached.Resp, req) }()
```
jyotimahapatra, line 156: do we have to return here?

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,183 @@ func New(ctx context.Context,
...
+	}
+
+	// Check if we have a upstream stream open for this aggregated key. If not,
+	// open a stream with the representative request.
+	//
+	// Locking is necessary here so that a simultaneous downstream request
+	// that maps to the same aggregated key doesn't result in two upstream
+	// streams.
+	o.upstreamResponseMap.mu.Lock()
```
jyotimahapatra, line 165: can we use a more granular lock per stream/key?

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,183 @@ func New(ctx context.Context,
...
+	shutdown func(),
+) {
+	for {
+		select {
+		case x, more := <-responseChannel:
+			if !more {
+				// A problem occurred fetching the response upstream, log and
+				// return the most recent cached response, so that the
+				// downstream will reissue the discovery request.
+				o.logger.With("key", aggregatedKey).Error(ctx, "upstream error")
```
jyotimahapatra, line 206: i think we should retry or return here.

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,183 @@ func New(ctx context.Context,
...
+			} else {
```
jyotimahapatra, line 207: the else block can perhaps be dropped. The code would still execute.

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,183 @@ func New(ctx context.Context,
...
+				// Cache the response.
```
jyotimahapatra, line 208: todo for handling this case better?

internal/app/orchestrator/orchestrator.go
jyotimahapatra, line 206: todo with gh issue for retrying?

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,183 @@ func New(ctx context.Context,
...
+			cached, err := o.cache.Fetch(aggregatedKey)
+			if err != nil {
+				o.logger.With("err", err).With("key", aggregatedKey).Error(ctx, "cache fetch failed")
+				// Can't do anything because we don't know who the watches
+				// are. Drop the response.
+			} else {
+				if cached == nil || cached.Resp == nil {
+					// If cache is empty, there is nothing to fan out.
+					if !more {
+						// Warn. Benefit of the doubt that this is the first request.
```
jyotimahapatra, line 230: since this block is inside select, there will always be a response, no?

internal/app/orchestrator/orchestrator.go
jessicayuen: Nope, because we still have to open an upstream stream if one doesn't exist for any multitude of reasons. There's no guarantee there's an open stream.

internal/app/orchestrator/orchestrator.go
lita: If the channel is closed, you will continue to get the default struct and more being false.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: right. got it.

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,183 @@ func New(ctx context.Context,
...
+			}
+		}(watch)
+	}
+}
+
+// onCacheEvicted is called when the cache evicts a response due to TTL or
+// other reasons. When this happens, we need to clean up open streams.
+// We shut down both the downstream watches and the upstream stream.
+func (o *orchestrator) onCacheEvicted(key string, resource cache.Resource) {
+	o.downstreamResponseMap.deleteAll(resource.Requests)
```
jyotimahapatra, line 267: i think we discussed that removing all sidecars will cause a thundering herd problem. We need a way to open a new stream on upstream only when ttl expires.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: yeah my point was when more is false, we should return because the channel will never return further.

internal/app/orchestrator/orchestrator.go
jessicayuen: I have the retry TODO set here https://github.com/envoyproxy/xds-relay/pull/53/files#diff-2958638164db92742469d2bb08d0001aR169-R171. 

Rather than issuing the retry in multiple places, I'm thinking if the upstream channel is closed (`!more`), we should just close all downstreams, and restart the cycle. Thoughts?

internal/app/orchestrator/orchestrator.go
jessicayuen: Chatted offline. We're going to add a TODO to issue a retry on the upstream rather than closing the downstream connections.

internal/app/orchestrator/orchestrator.go
jessicayuen: Talked offline, will add the TODO here as well.

internal/app/orchestrator/orchestrator.go
jessicayuen: opened https://github.com/envoyproxy/xds-relay/issues/71

internal/app/orchestrator/orchestrator.go
jessicayuen: I don't think we can use a more granular lock because we're modifying the map to add keys further down https://github.com/envoyproxy/xds-relay/pull/53/files#diff-2958638164db92742469d2bb08d0001aR173

jyotimahapatra: All simultaneous requests will be blocking on the slowest OpenStream, isnt it?

jessicayuen: I moved to `sync.Map` for per-key concurrency, let me know your thoughts: https://github.com/envoyproxy/xds-relay/pull/53/commits/aeef7ca998b42f1d2fe9237d17186003a397f91b

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,186 @@ func New(ctx context.Context,
...
-func (c *orchestrator) onCacheEvicted(key string, resource cache.Resource) {
-	// TODO implement.
+// watchResponse is intended to be called in a goroutine, to receive incoming
+// responses and fan out to downstream clients.
+func (o *orchestrator) watchUpstream(
+	ctx context.Context,
+	aggregatedKey string,
+	responseChannel <-chan *discovery.DiscoveryResponse,
+	done <-chan bool,
+	shutdown func(),
```
jyotimahapatra, line 198: nit: s/shutdown/shutdownUpstream/

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,186 @@ func New(ctx context.Context,
...
+			shutdown()
+			return
+		}
+	}
+}
+
+// fanout pushes the response to the response channels of all open downstream
+// watches in parallel.
+func (o *orchestrator) fanout(resp *cache.Response, watchers map[*gcp.Request]bool) {
+	for watch := range watchers {
```
jyotimahapatra, line 255: probably need to make a copy and then iterate?

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,186 @@ func New(ctx context.Context,
...
+		go func(watch *gcp.Request) {
+			if channel, ok := o.downstreamResponseMap.get(watch); ok {
+				channel <- convertToGcpResponse(resp, *watch)
```
jyotimahapatra, line 258: need a default case?

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +114,186 @@ func New(ctx context.Context,
...
+			return
+		}
+	}
+}
+
+// fanout pushes the response to the response channels of all open downstream
+// watches in parallel.
+func (o *orchestrator) fanout(resp *cache.Response, watchers map[*gcp.Request]bool) {
+	for watch := range watchers {
+		go func(watch *gcp.Request) {
```
jyotimahapatra, line 256: this will cause a temporary spike in goroutines, are the goroutines needed?

internal/app/orchestrator/orchestrator.go
```diff
@@ -56,31 +64,42 @@ type orchestrator struct {
...
+		upstreamResponseMap:   newUpstreamResponseMap(),
 	}
 
 	// Initialize cache.
-	cache, err := cache.NewCache(int(cacheConfig.MaxEntries),
+	cache, err := cache.NewCache(
+		int(cacheConfig.MaxEntries),
 		orchestrator.onCacheEvicted,
-		time.Duration(cacheConfig.Ttl.Nanos)*time.Nanosecond)
+		time.Duration(cacheConfig.Ttl.Nanos)*time.Nanosecond,
```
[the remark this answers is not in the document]
jessicayuen: Yeah this wasn't added as a part of this PR. Just a format change, we can open an issue but I wouldn't link it as a part of this.

internal/app/orchestrator/orchestrator.go
jessicayuen: Are you worried about cache ttl removing the watches in the middle of the fanout? I'm afraid we can't make a copy as the downstreamResponseMap requires a pointer to the watch, in order to retrieve the channel.

internal/app/orchestrator/orchestrator.go
jessicayuen: This was a suggestion [here](https://github.com/envoyproxy/xds-relay/pull/53#discussion_r407537441). Ultimately I decided to go with this so we're not blocking on the slowest watch, if there are many. Pushing to the channel is quick but there may be some overhead in the conversion to gcp response.

internal/app/orchestrator/orchestrator.go
jessicayuen: Is your proposal to remove it from the cache if we can't find a corresponding channel for the watch? This shouldn't happen if we're cleaning up properly but I can add the sanity check.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: there can be multiple situations in which the watch could be changed.  a watch could be removed when sidecar disconnects and also on ttl expiry. Im not sure how iterators in java work. in java changing a iterator while changing it will cause exception.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: the channel could be blocked for many reasons. if its blocked, we will need a default case, else the whole flow will be stuck on slowest response.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: therre could be 1000s of watches per key. I am concerrned about the implications of starting so many goroutines.

internal/app/orchestrator/orchestrator.go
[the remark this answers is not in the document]
jyotimahapatra: sounds good.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: since you are using gororutines, it wont be a problem, but imagine one slow channel. There will be multiple goroutines backed up to ssend response. We wont be able to guarentee which one goes first. If teh goroutines are rremoved, then default case will be necessary.

internal/app/orchestrator/orchestrator.go
jessicayuen: I'm not too concerned about the burst in routines. Someone on stack overflow did an experiment on the memory usage and time to start a go routine on 100k go routines, https://stackoverflow.com/questions/8509152/max-number-of-goroutines

```
Go 1.6.2 (April 2016)
  32-bit x86 CPU (A10-7850K 4GHz)
    | Number of goroutines: 100000
    | Per goroutine:
    |   Memory: 4536.84 bytes
    |   Time:   1.634248 µs
  64-bit x86 CPU (A10-7850K 4GHz)
    | Number of goroutines: 100000
    | Per goroutine:
    |   Memory: 4707.92 bytes
    |   Time:   1.842097 µs
```

Ends up being about 0.5gb of memory for 100k routines. I don't anticipate a single xds relay instance having more watches than than that (or anywhere close tbh). If this ends up being problematic, I would rather revisit and remove the go routine then. Thoughts?

internal/app/orchestrator/orchestrator.go
jessicayuen: I put a [comment below](https://github.com/envoyproxy/xds-relay/pull/53#discussion_r417697540) regarding the routines but I'll still add a default case. Maybe to timeout and log error if we're still not able to push to it after 10s?

jyotimahapatra: sounds good. Could use the `util.DoWithTimeout` for the purpose.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: one aspect im thinking is when we schedule many goroutines, there is no guarentee that they will be scheduled in order. How do we make sure the older data does not override ea more recent update ?

internal/app/orchestrator/orchestrator.go
jessicayuen: This should be fine in Go: https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop

internal/app/orchestrator/orchestrator.go
jessicayuen: How about if I add a WaitGroup to await on this fanout completing (all go routines complete) before the function exits?

internal/app/orchestrator/orchestrator.go
jyotimahapatra: when the watch is deleted, the channel will be gone too. We will be pushing into a closed channel, isnt it?

internal/app/orchestrator/orchestrator.go
jessicayuen: Synced offline. We decided to use the WaitGroup, drop the timeout, just add a default log case. This way we don't block on future responses for the duration of the slowest sidecar.

internal/app/orchestrator/orchestrator.go
jessicayuen: Synced offline. Our decision is to not close the downstream channels (on cache evict and on cancel). Garbage collection will clean up and close the channels. This way we don't have to worry about pushing to a closed channel in a different routine.

internal/app/orchestrator/orchestrator.go
jessicayuen: @jyotimahapatra I reduced the timeout to 1s rather than changing it to a default. When I changed it to default, the test blocks [here](https://github.com/envoyproxy/xds-relay/pull/53/files#diff-2c2ed0f611d2764f6a076c08ee2b18f7R309).

jyotimahapatra: Do you know why the test blocked? imo we shoud put default instead of a timeout. A timeout is not guaranteed to help because Send blocks when transport is blocked or stream is broken.
For handling safe delivery we need to design some way to retry more systematically.

jessicayuen: The downstream channels were unbuffered so the `default` case was always being selected in the test run. I changed these to buffered channels and it seems to be fine now.

jessicayuen:
@jyotimahapatra Addressed all comments, can you take another pass?

internal/app/orchestrator/orchestrator_test.go
```diff
@@ -44,3 +123,216 @@ func TestNew(t *testing.T) {
...
+
+	upstreamResponseChannelLDS <- &upstreamResponseLDS
+	upstreamResponseChannelCDS <- &upstreamResponseCDS
+
+	gotResponseFromChannel1 := <-respChannel1
+	gotResponseFromChannel2 := <-respChannel2
+	gotResponseFromChannel3 := <-respChannel3
+
+	assert.Equal(t, 3, len(orchestrator.downstreamResponseMap.responseChannels))
+	testutils.AssertSyncMapLen(t, 2, orchestrator.upstreamResponseMap.internal)
```
lita, line 323: Is asserting the contents of the channel good enough vs actually checking the keys?

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +115,197 @@ func New(ctx context.Context,
...
+// watchResponse is intended to be called in a goroutine, to receive incoming
+// responses and fan out to downstream clients.
+func (o *orchestrator) watchUpstream(
+	ctx context.Context,
+	aggregatedKey string,
+	responseChannel <-chan *discovery.DiscoveryResponse,
+	done <-chan bool,
+	shutdownUpstream func(),
+) {
+	for {
```
lita, line 201: This logic really hurts my head. It would be great to add a comment here about how it listens for responses from the upstream channel, caches the response  and fans out to the change to the downstream channel.

It would also be good to document how this shuts down and the purpose of the shutdownUpstream function. I assume it is via the done channel.

internal/app/orchestrator/orchestrator.go
```diff
@@ -95,16 +115,197 @@ func New(ctx context.Context,
...
+	var wg sync.WaitGroup
+	for watch := range watchers {
+		wg.Add(1)
+		go func(watch *gcp.Request) {
+			defer wg.Done()
+			if channel, ok := o.downstreamResponseMap.get(watch); ok {
+				select {
+				case channel <- convertToGcpResponse(resp, *watch):
+					break
+				default:
```
lita, line 265: How comment is this expected to happen? It looks like this channel is buffered as 1. if there is spam of updates, maybe it is fine to drop?

internal/app/orchestrator/orchestrator.go
jessicayuen: > Is there a reason why the map is a pointer of the request rather than value?

Yes, because value isn't sufficient to get the corresponding downstream watch. Cache was originally by value, and I was not able to retrieve the same watch so we moved all these to pointers.

internal/app/orchestrator/upstream.go
```diff
@@ -20,45 +26,48 @@ type upstreamResponseChannel struct {
...
-) upstreamResponseChannel {
+) (upstreamResponseChannel, bool) {
 	channel := upstreamResponseChannel{
 		response: responseChannel,
 		done:     make(chan bool, 1),
 	}
-	u.responseChannels[aggregatedKey] = channel
-	return channel
+	result, exists := u.internal.LoadOrStore(aggregatedKey, channel)
+	return result.(upstreamResponseChannel), exists
```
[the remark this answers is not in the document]
jessicayuen: Yeah, `internal` should never be accessed outside of this file so this is fine.

internal/app/orchestrator/orchestrator_test.go
jessicayuen: I primarily cared about the length since I do value checking below when the content flows downstream:
```
assertEqualResources(t, gotResponseFromChannel1, upstreamResponseLDS, req1)
assertEqualResources(t, gotResponseFromChannel2, upstreamResponseLDS, req2)
assertEqualResources(t, gotResponseFromChannel3, upstreamResponseCDS, req3)
```
but I agree that a key/value check will be a more complete test. I'll make the change.

internal/app/orchestrator/downstream.go
```diff
@@ -0,0 +1,69 @@
+package orchestrator
```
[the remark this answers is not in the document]
jessicayuen: sg. I'll also add one for the upstream.

internal/app/cache/cache.go
```diff
@@ -193,7 +194,9 @@ func (c *cache) getExpirationTime(currentTime time.Time) time.Time {
 	return time.Time{}
 }
 
-func marshalResources(resources []*any.Any) ([]gcp_types.MarshaledResource, error) {
+// MarshalResource converts the raw xDS discovery resources into a serialized
+// form accepted by go-control-plane.
+func MarshalResources(resources []*any.Any) ([]gcp_types.MarshaledResource, error) {
```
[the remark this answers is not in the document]
jessicayuen: I think that's a good suggestion. Opened an issue for this https://github.com/envoyproxy/xds-relay/issues/79

lita:
Overall, this looks great. Most of my comments are around documentation the code, but the logic of the implementation makes sense to me.

internal/app/orchestrator/orchestrator.go
jessicayuen: oh no :( 

I do document the first part at the [function header](https://github.com/envoyproxy/xds-relay/pull/53/files#diff-2958638164db92742469d2bb08d0001aR192-R193) and make comments throughout, although the documentation for done/shutdownUpstream could be more clear. I'll put more details in the function header comment.

internal/app/orchestrator/orchestrator.go
jessicayuen: There's more discussion for this [here](https://github.com/envoyproxy/xds-relay/pull/53/files#r417671109).

The idea behind this was that a downstream receiver could be blocked for any reason. Initially I had went with a timeout, but in discussions with @jyotimahapatra we realized that it would imply the slowest sidecar would block new updates from being sent to other sidecars, due to the use of WaitGroups. We can't get rid of WaitGroups because more recent updates might get overriden by a slower update if updates were bursty.

The buffer is necessary in order to give time for the receiver to initialize. I realized in [this test](https://github.com/envoyproxy/xds-relay/pull/53/files#diff-2c2ed0f611d2764f6a076c08ee2b18f7R254) that the second watch (for the same aggregated key) would fall to the `default` case if the channel was unbuffered.

If there is a spam of updates, I think it's reasonable to drop since only one of those updates will be applied downstream anyway, but we do lose the the guarantee that it is the "latest" update. Could move back to the configured timeout model if this is a concern, but I put some of those downsides ^.

jessicayuen: My thoughts are to move forward with the default for now. I'll document this. If it is a problem in practice, then we can revisit the timeout model or other approaches.

jyotimahapatra: sounds good to me.

lita: Yeah in the future, you can do what Envoymanager does and drop the oldest and enqueue the newest. That has been working quite well for us.

jyotimahapatra:
@jessicayuen The changes look great. I was thinking there's scope for adding some negative tests too for the orchestrator. If you want to add those in a separate PR, i would be ok with that.

jessicayuen:
Sounds reasonable @jyotimahapatra, let me add that in a separate PR or as a part of the e2e tests - this PR is already getting a bit unwieldy and I would like to get a base point in.

gnz00:
@lita I might be misunderstanding, there are some ordering concerns with ADS. There has been a few issues about it on the go-control-plane repo (one of them from myself). 

@kyessenov summarizes it best, union updates and version transactions: https://github.com/envoyproxy/go-control-plane/issues/218#issuecomment-527682986

PR #54, opened 2020-04-10
Alternate xdsimpl

jyotimahapatra:
This is an alternate approach of xdsimpl where the error is retried inside xdsimpl. The error is not passed to the caller.


internal/app/upstream/client.go
```diff
@@ -55,11 +79,206 @@ type Response struct {
...
+		close(response)
+		return nil, err
+	}
+
+	continueChan := make(chan bool, 1)
+	continueChan <- true
+	go func() {
+		for {
+			select {
+			case <-continueChan:
```
jyotimahapatra, line 117: @jessicayuen This is the alternate approach we discussed about. The error is not sent back to the caller and retried continuously. There might be other ways to achieve the same, but this approach is as good as others.

PR #59, opened 2020-04-16
DeleteRequest API

### LINKED ISSUE #57: Add `DeleteWatch` in Cache
When onCancel is called from go-control-plane to the Orchestrator, Orchestrator removes management of this watch, but the watch persists in the Cache until the TTL expires. We could more effectively manage resources if a DeleteWatch(req *v2.DiscoveryRequest) function existed.

LisaLudique:
Fixes #57 

This change adds an API to the cache to support deleting a request when given the key and the (pointer to the) request, which the orchestrator can use upon cancellation. If the key is not in the cache, nothing happens.

The input type to AddRequest was also changed to take a pointer to a request.


internal/app/cache/cache.go
```diff
@@ -185,3 +204,16 @@ func marshalResources(resources []*any.Any) ([]gcp_types.MarshaledResource, erro
...
 }
+
+func removeRequests(request *v2.DiscoveryRequest, requests []*v2.DiscoveryRequest) []*v2.DiscoveryRequest {
+	requestsAfterDeletion := make([]*v2.DiscoveryRequest, len(requests))
+	i := 0
+	for _, r := range requests {
+		if r != request {
+			requestsAfterDeletion[i] = r
+			i++
+		}
```
jyotimahapatra, line 215: I'm thinking this loop could be very long. Wondering if there was a different way.

LisaLudique: I thought about using a map to track all the requests instead, but there's still a tradeoff in efficiency. Using a list like we do today makes AddRequest and SetResponse more efficient and DeleteRequest less efficient w.r.t requests. Using a map would be AddRequest and DeleteRequest more efficient but SetResponse less efficient (need to iterate over the map to get the keys), so it depends on whether we expect to call SetResponse or DeleteRequest more often.

jyotimahapatra: I think the frequency of Add/Delete Requests is much more than SetResponse.
The scenario i'm thinking about is, say, during scaleup 1000 pods came up. When they get removed during scaledown 1000 sidecar DeleteRequest will cause 1000^2 iterations.

LisaLudique: As discussed, since the caller of SetResponse would iterate over the return value (list of requests) anyway, it should make no difference to return a map of requests instead for the caller to iterate over.

LisaLudique: The map value type could also be an empty struct if the added memory from carrying around an extra bool is too much, but that might be excessively optimizing.

internal/app/cache/cache.go
```diff
@@ -19,10 +19,13 @@ type Cache interface {
 	Fetch(key string) (*Resource, error)
 
 	// SetResponse sets the cache response and returns the list of requests.
-	SetResponse(key string, resp v2.DiscoveryResponse) ([]*v2.DiscoveryRequest, error)
+	SetResponse(key string, resp v2.DiscoveryResponse) (map[*v2.DiscoveryRequest]bool, error)
```
jessicayuen, line 22: Just a thought ...
Alternative is to keep this interface and put the map into the [internal cache struct](https://github.com/envoyproxy/xds-relay/pull/59/files#diff-daf2807d54a5369eb2a29014b0d051cfR31). An advantage of keeping the slice is ordered keys when orchestrator or other callers wants to test the results against a mocked Response. But sorting adds an overhead to SetResponse. Since orchestrator doesn't care about order during fanout, this is probably ok or preferable for now.

PR #62, opened 2020-04-20
End-to-end tests

eapolinario:
This PR adds support for end-to-end tests involving `go-control-plane` SnapshotCache as the backbone for a management server, `xds-relay` server and one instance of envoy.

The setup is similar to how `go-control-plane` does its [e2e tests](https://github.com/envoyproxy/go-control-plane/blob/master/pkg/test/main/main.go), although in our case we decided to adopt a more conventional test structure, the idea being that we're going to extend the number of tests as time goes by.

Speaking about the mechanics of the tests, we run them in a container to make it easy to bring a real instance of envoy. By choosing to do it all in golang we had to sacrifice portability since there is no cross-platform way of safely exiting processes spawned in go. I ran multiple tests with more than one test in the same suite binding to the same ports (basically duplicates of the tests, only with different names) and had to introduce a 1 second delay between tests to make sure the OS can free up the ports between tests.

I also left a few TODO, but I'd like to get feedback on the overall approach.


integration/e2e_test.go
```diff
@@ -0,0 +1,230 @@
...
+func TestMain(m *testing.M) {
+	// We force a 1 second sleep before running a test to let the OS close any lingering socket from previous
+	// tests.
+	time.Sleep(1 * time.Second)
+	code := m.Run()
+	os.Exit(code)
+}
+
+func TestSnapshotCacheSingleEnvoyAndXdsRelayServer(t *testing.T) {
+	if runtime.GOOS != "linux" {
```
jyotimahapatra, line 41: since there are many nuances for running this and envoy is the only out of process component, would it be better to make this a bash driven test?

.github/workflows/end2end-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          path: ${{ env.PACKAGEPATH }}
+      - name: Build image
+        working-directory: ${{ env.PACKAGEPATH }}
+        run: docker build . --file Dockerfile --tag xdsrelay
```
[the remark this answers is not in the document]
eapolinario: Good question. The documentation doesn't mention anything, I'll see what I can find out.

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM golang:1.14.1
+
+WORKDIR /xds-relay
+
+# add envoy
+COPY --from=envoyproxy/envoy:v1.13.1 /usr/local/bin/envoy /usr/local/bin/envoy
```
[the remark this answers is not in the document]
eapolinario: We could probably use the [`envoyproxy/envoy-dev`](https://hub.docker.com/r/envoyproxy/envoy-dev/tags/) image, although we'd have to discuss how to deal with failures / tags, etc.

integration/e2e_test.go
eapolinario: I thought about it and decided against a bash-driver test because I was envisioning a scenario where we control multiple envoy instances from the test. Imagine being able to stand up and down envoy instances in the middle of a test.

eapolinario: Actually, I was looking around the solution used by gcp (namely trap the  even if made this a bash-driven test we would still suffer from the same problem (i.e. no cross-platform way of killing child processes and its descendants). The solution adopted in gcp (i.e. trapping the exit of the bash process to kill the parent envoy process) only kills the parent process. What I'm getting at is that even if we moved to bash we would still need to rely on a linux-only solution to handle this. 

Another alternative I considered was to try to make sure that ports were unique in each test, but that also brings its own set of complications and was quickly shot down by me.

.github/workflows/end2end-tests.yml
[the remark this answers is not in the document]
eapolinario: This is not documented anywhere, but I tested this on a private repo of my own and the generated docker image does not count against my storage quota. I don't have permissions to check the usage of the `xds-relay` project, but I'm going to assume that it's the same (since we're not really generating any egress calls to an outside registry or anything).

Dockerfile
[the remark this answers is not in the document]
jyotimahapatra: yeah..this is not an issue for now

.github/workflows/end2end-tests.yml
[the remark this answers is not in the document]
jyotimahapatra: sounds good. We can cleanup if needed later.

integration/e2e_test.go
jyotimahapatra: `Imagine being able to stand up and down envoy instances in the middle of a test` this is definitely a compelling scenario for avoiding a bash driver.

integration/e2e_test.go
```diff
@@ -0,0 +1,230 @@
...
+	defer cancelFunc()
+
+	// Test parameters
+	var port uint = 18000
+	var upstreamPort uint = 18080
+	var basePort uint = 9000
+	var nClusters = 7
+	var nListeners = 9
+	var nUpdates = 4
+
```
jyotimahapatra, line 57: nit: may be keep these in a `const` block to keep the surface area of the test smaller.

integration/e2e_test.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+
+		// TODO: parametrize node-id in bootstrap files.
+		err := configv2.SetSnapshot("test-id", snapshotv2)
+		if err != nil {
+			t.Fatalf("Snapshot error %q for %+v\n", err, snapshotv2)
+		}
+
+		g.Eventually(func() (int, int) {
+			ok, failed := callLocalService(basePort, nListeners)
```
jyotimahapatra, line 89: we can probably avoid setting up and calling a local service and instead use envoy stats to check config was applied. That way we can probably remove dependency on `Eventually`. wdyt?

eapolinario: we'd still have a similar problem, right? As in, at which point can we check the `/stats` enpdoint?

jyotimahapatra: yeah i see..`Eventually` still has to be used. can we still use envoy stats to verify instead of a local service.

jyotimahapatra: as i think more, i think what we have is better since it validates lds/cds/rds/eds were configured right.

integration/testdata/envoy_bootstrap.yaml
```diff
@@ -0,0 +1,39 @@
...
+      address: 127.0.0.1
+      port_value: 19000
+dynamic_resources:
+  cds_config:
+    api_config_source:
+      api_type: GRPC
+      grpc_services:
+      - envoy_grpc:
+          cluster_name: xds_cluster
+      set_node_on_first_message_only: true
```
jyotimahapatra, line 15: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/config_source.proto In our design we always expect xds relay to have this `set_node_on_first_message_only` to be false?

jyotimahapatra: @jessicayuen wdyt about the constraint in our system?

jyotimahapatra: its not a concern since gcp handles it https://github.com/envoyproxy/go-control-plane/blob/master/pkg/server/v2/server.go#L301-L305

integration/e2e_test.go
jessicayuen: @eapolinario Out of curiosity, did you test this locally on a linux system then?

jyotimahapatra: I run my g-c-p ingetration tests like this `docker run -v ~/src/go-control-plane:/go-control-plane --cpus=10 -it gcr.io/istio-testing/go-control-plane-ci:04-02-2020 /bin/bash`

Makefile
```diff
@@ -20,6 +20,10 @@ unit: ## Run all unit tests with coverage report
 integration-tests:  ## Run integration tests
 	go test -tags integration -v ./integration/
 
+.PHONY: e2e-tests
+e2e-tests: ## Run e2e tests
```
jessicayuen, line 24: should this have a dependency on `build-docker-image` to create an image if it doesn't exist?

eapolinario: how about we check if the image is built in the `/run_docker.sh` script instead? Otherwise that script will not be super useful to run the tests locally.

integration/e2e_test.go
eapolinario: @jessicayuen , I added a script to help run these tests on a mac. Check the `run_docker.sh` file. Assuming you have the docker runtime installed on your mac (super easy to install via homebrew: `brew cask install docker`), you can run this locally by running `./scripts/run_docker.sh make e2e-tests`.

PR #63, opened 2020-04-20
xdsclient integration tests

jyotimahapatra:
The xdsimpl integration tests doesn't need a envoy sidecar and can be written as integration tests rather than [e2e tests](https://github.com/envoyproxy/xds-relay/pull/62).

The integration tests help in strengthening the grpc mock behaviors in the upstream client unit tests.

The tests use g-c-p server and xdsclient interacting over real grpc.


integration/client_test.go
```diff
@@ -0,0 +1,295 @@
...
+	defer cancel()
+	testLogger := log.New(loglevel)
+
+	snapshotsv2, configv2 := createSnapshotCache(updates, testLogger)
+	cb := gcptestv2.Callbacks{Signal: make(chan struct{})}
+	respCh, shutdown, err := setup(ctx, testLogger, snapshotsv2, configv2, &cb)
+	if err != nil {
+		assert.Fail(t, "Setup failed: %s", err.Error())
+		return
+	}
```
jessicayuen, line 93: Do you want to pull this out to a helper / setup function since it's used across a few of these tests?

integration/client_test.go
```diff
@@ -0,0 +1,295 @@
...
+func createSnapshotCache(updates int, logger log.Logger) (resourcev2.TestSnapshot, gcpcachev2.SnapshotCache) {
+	return resourcev2.TestSnapshot{
+		Xds:          "xds",
+		UpstreamPort: 18080,
+		BasePort:     9000,
+		NumClusters:  updates,
+	}, gcpcachev2.NewSnapshotCache(false, gcpcachev2.IDHash{}, gcpLogger{logger: logger})
+}
+
+type gcpLogger struct {
```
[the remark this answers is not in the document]
jyotimahapatra: yes..we talked about it this morning. We can merge these PRs in and I can remove the duplications in a separate PR. There are other parts which could be unified as well.

integration/client_test.go
jyotimahapatra: I think i have pulled the most of the common code into setup. I did not want to return 5 things out of setup and hence the repeated lines
```
        snapshotsv2, configv2 := createSnapshotCache(updates, testLogger)
	cb := gcptestv2.Callbacks{Signal: make(chan struct{})}
	respCh, shutdown, err := setup(ctx, testLogger, snapshotsv2, configv2, &cb)
```

integration/client_test.go
```diff
@@ -0,0 +1,295 @@
+// +build integration
+
```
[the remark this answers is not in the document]
jyotimahapatra: done.

integration/client_test.go
jyotimahapatra: I have moved the logger out to a common place. I think the other common line is ctx creation, which is specific to each test.

integration/client_test.go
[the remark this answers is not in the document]
jessicayuen: Sounds good, I thought this morning's discussion was related to the snapshot cache.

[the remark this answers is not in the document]
jyotimahapatra: since both are parallel PRs, we can merge them and i can quickly remove the duplications in a new PR when both are merged.

integration/upstream_client_test.go
```diff
@@ -0,0 +1,291 @@
...
+				version++
+				wg.Done()
+			}
+		}
+	}()
+
+	sendResponses(ctx, testLogger, updates, snapshotsv2, configv2)
+	wg.Wait()
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
jessicayuen, line 71: Since you're reassigning cancel here, does the [first one](https://github.com/envoyproxy/xds-relay/pull/63/files#diff-87e47ce072c022401a8b05fe9def3110R41) get properly cleaned up?

jyotimahapatra: hm i didnt notice i was reusing the variable...i will put unique names to avoid confusion.

jyotimahapatra: Thanks for catching. done.

integration/client_test.go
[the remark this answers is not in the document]
jyotimahapatra: since the e2e test PR is now merged, i've reused the logger now.

jyotimahapatra:
@jessicayuen i have removed some common code now. PTAL

integration/e2e_test.go
```diff
@@ -85,27 +88,27 @@ func TestSnapshotCacheSingleEnvoyAndXdsRelayServer(t *testing.T) {
 		// TODO: parametrize node-id in bootstrap files.
 		err := configv2.SetSnapshot("test-id", snapshotv2)
 		if err != nil {
-			t.Fatalf("Snapshot error %q for %+v\n", err, snapshotv2)
```
jessicayuen, line 88: Will logger.Error cause the test to terminate early? If not, I think we may need to leave the Fatals alone.

jessicayuen: The Logger Interface also has a Fatal function, I think you can implement that and use it instead of Error.

jyotimahapatra: i think its a fair thing to terminate early when error happens , no?

jyotimahapatra: changed them to fatal now

PR #64, opened 2020-04-22
grpc graceful shutdown

jyotimahapatra:


integration/server_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	binaryName = "xds-relay"
+)
+
+func TestXdsClientGetsIncrementalResponsesFromUpstreamServer(t *testing.T) {
+	dir, err := os.Getwd()
```
lita, line 21: Is this guaranteed to be here in CI jobs? Should there be an environment variable but default to this if it is not set?

internal/app/upstream/client.go
```diff
@@ -79,6 +79,7 @@ type version struct {
 // The method does not block until the underlying connection is up.
 // Returns immediately and connecting the server happens in background
 func NewClient(ctx context.Context, url string, callOptions CallOptions, logger log.Logger) (Client, error) {
+	logger.Debug(ctx, "Initiating upstream connection.")
```
lita, line 82: Should this be info?

internal/app/server/server.go
```diff
@@ -49,52 +64,41 @@ func Run(bootstrapConfig *bootstrapv1.Bootstrap,
...
+func registerShutdownHandler(server *grpc.Server, logger log.Logger) {
 	sigs := make(chan os.Signal, 1)
+	ctx := context.Background()
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 	go func() {
-		<-sigs
-		_ = util.DoWithTimeout(context.Background(), func() error {
+		sig := <-sigs
+		logger.Info(ctx, "received interrupt signal:", sig.String())
+		err := util.DoWithTimeout(ctx, func() error {
```
lita, line 95: Would it be possible to pass in a Timeout via WithTimeout through the context? That way, we can add tests when the timeout was hit and we did not gracefully shutdown.

integration/server_test.go
jyotimahapatra: yeah..the CI jobs are set up to correctly set the folder https://github.com/envoyproxy/xds-relay/blob/master/.github/workflows/integration-tests.yml#L40
They run locally too without any changes.

internal/app/upstream/client.go
jyotimahapatra: Yea makes sense. Will do.

internal/app/server/server.go
jyotimahapatra: I think passing a timeout via WithTimeout through the context could lead to bugs. Usually the `ctx` param in any function is the application context. Putting the notion of passing in a timeout context in param could lead to bugs.
Also the `DoWithTimeout` function is already unit tested for all scenarios it serves.

Since the code deals with system signals i felt it would be hard to unit test unless the system signals and grpc methods are mocked out. I felt it would lead to complicated indirections. I thought about trying to make `registerShutdownHandler` a utility function and try to mock and unit test the negative case, but i didn;t pursue it because the shutdown handler is not supposed to be a utility and not supposed to be used anywhere else in the code.

internal/app/server/server.go
```diff
@@ -36,6 +40,13 @@ func Run(bootstrapConfig *bootstrapv1.Bootstrap,
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// Initialize request aggregation mapper component.
+	requestMapper := mapper.NewMapper(aggregationRulesConfig)
```
jessicayuen, line 44: nit: I wouldn't move this up here so we can just exit on `if mode != "serve"`

internal/app/server/server.go
```diff
@@ -36,6 +40,13 @@ func Run(bootstrapConfig *bootstrapv1.Bootstrap,
 	ctx, cancel := context.WithCancel(context.Background())
```
jessicayuen, line 40: Should you be passing this ctx and cancel pair to the shutdown handler?

internal/app/server/server.go
```diff
@@ -72,10 +79,27 @@ func Run(bootstrapConfig *bootstrapv1.Bootstrap,
...
+
+func registerShutdownHandler(server *grpc.Server, logger log.Logger) {
+	sigs := make(chan os.Signal, 1)
+	ctx := context.Background()
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		sig := <-sigs
+		logger.Info(ctx, "received interrupt signal:", sig.String())
+		err := util.DoWithTimeout(ctx, func() error {
+			logger.Info(ctx, "initiating grpc graceful stop")
```
jessicayuen, line 97: Additionally, I think you should flush any outstanding logs, by calling `logger.Sync()`

jyotimahapatra: i see..will add that here.

internal/app/server/server.go
jyotimahapatra: right. will do that.

jyotimahapatra: actually there;s a `defer cancel()` already. https://github.com/envoyproxy/xds-relay/blob/master/internal/app/server/server.go#L37
We need not cancel the ctx in shutdown handler.

internal/app/server/server.go
lita: Yeah but I think it's important to test the behavior that everything does shutdown, whether the client/server is misbehaving.

You can set the WithCancel or WithTimeout as an application variable so that the timeout is pretty short. I also figured you want the timeout to be configurable in some way.

internal/app/server/server.go
jyotimahapatra: i reverted the placement to where it was before to keep the scope of PR limited.

internal/app/server/server.go
jyotimahapatra: Yeah we should make the timeout configurable. Will do it.

About the test, I would like to brainstorm a bit of the approach you are thinking. Are you suggesting to continue using the integration test, have a client connect to xds relay and then signal xds-relay to shutdown. Even if we do these with a short timeout, there is no way to guarantee that graceful shutdown wont happen faster than that, since we dont control grpc stack.
Or are you suggesting to use unit tests to validate the behaviour?

internal/app/server/server.go
jessicayuen: In this case, can you get rid of the above TODO (I know you didn't add it)? I added it initially thinking that the function may not exit properly and cancel never gets called.

jyotimahapatra: i thought some more and figured there's value in cancelling the ctx anyway. Thanks for suggesting.

internal/app/server/server.go
jyotimahapatra: I have added a unit test. @lita lmk if the test will work for the scenario.

PR #74, opened 2020-05-01
Admin Server and configdump endpoint

LisaLudique:
This change sets up the admin server and the `server_info` endpoint #77 
It also implements part of the `cache` endpoint by returning the cache resource corresponding to the given key #75 but does not yet implement a full cache dump or querying by regex.

Example outputs after running `go run main.go -a integration/testdata/keyer_configuration_complete_tech_spec.yaml -c integration/testdata/bootstrap_configuration_complete_tech_spec.yaml -l debug -m serve`:

Default handler:
```
➜  xds-relay git:(admin) ✗ curl 127.0.0.1:6070/                 
admin commands are:
  /: admin home page
  /cache/: print cache entry for a given key. usage: `/cache/<key>`
  /server_info: print bootstrap configuration
```

Cache dump handler:
```
➜  xds-relay git:(admin) curl 127.0.0.1:6070/cache/
no resource for key found in cache.    
```

Config dump handler:
```
➜  xds-relay git:(admin) curl 127.0.0.1:6070/server_info
{
  "server": {
    "address": {
      "address": "127.0.0.1",
      "port_value": 9991
    }
  },
  "origin_server": {
    "address": {
      "address": "my-control-plane.lyft.net",
      "port_value": 80
    }
  },
  "logging": {
    "path": "/var/log/xds-aggregator",
    "level": 1
  },
  "cache": {
    "ttl": {
      "seconds": 60
    },
    "max_entries": 10
  },
  "admin": {
    "address": {
      "address": "127.0.0.1",
      "port_value": 6070
    }
  }
}            
```


internal/app/orchestrator/orchestrator.go
```diff
@@ -98,12 +99,18 @@ func New(
 		orchestrator.logger.With("error", err).Panic(ctx, "failed to initialize cache")
 	}
 	orchestrator.cache = cache
+	http.HandleFunc("/cache/", orchestrator.CacheDumpHandler)
```
jessicayuen, line 102: What are your thoughts on moving the handlers to it's own package? `internal/app/http/handler` or whatnot? I haven't looked into the best practices for this, but I'd prefer to keep the orchestrator contained to the request/response management flow.

internal/app/server/server.go
```diff
@@ -44,6 +52,18 @@ func RunWithContext(ctx context.Context, cancel context.CancelFunc, bootstrapCon
 		logger = log.New(bootstrapConfig.Logging.Level.String())
 	}
 
+	// Configure admin server.
+	adminServer := &http.Server{
+		//TODO(lisalu): Make below address configurable.
```
jessicayuen, line 57: Do you plan to make this configurable as a part of this PR?

internal/app/server/server.go
```diff
@@ -44,6 +52,18 @@ func RunWithContext(ctx context.Context, cancel context.CancelFunc, bootstrapCon
...
+		Addr: "127.0.0.1:6070",
+	}
+
+	http.Handle("/", defaultHandler())
+	configDumpHandler := func(w http.ResponseWriter, req *http.Request) {
```
jessicayuen, line 62: Thoughts on moving this handler to the same package as mentioned above too?

internal/app/server/server.go
```diff
@@ -106,6 +129,9 @@ func registerShutdownHandler(
 			logger.Info(ctx, "initiating grpc graceful stop")
 			gracefulStop()
 			_ = logger.Sync()
+			if shutdownErr := adminShutdown(ctx); shutdownErr != nil {
+				logger.With("err", shutdownErr).Error(ctx, "admin shutdown error: ", shutdownErr.Error())
+			} // TODO(lisalu): Where to place admin shutdown?
```
jessicayuen, line 134: I think how you have it is good. I'd suggest shutting it down before `gracefulStop()` is called though, otherwise orchestrator and other modules may have cleaned up things required that the admin server still requires.

internal/app/orchestrator/orchestrator.go
LisaLudique: I was concerned about the accessbility of the resources (the cache specifically) but I'll see if I can move everything out into a separate package.

internal/app/server/server.go
LisaLudique: Yep

internal/app/orchestrator/orchestrator.go
LisaLudique: I moved things around in my latest commit, including the cache dump handler (and a first attempt at just doing the Fetch based on a key passed in). I had to create an accessor for the orchestrator cache. LMK what you think of this approach versus registering the handler in the orchestrator.

LisaLudique: One advantage of having this handler live with the orchestrator is that the orchestrator has access to all the cache keys (I think this is the case based on the discussion around DeleteWatch) so it makes implementing a full cache dump easier.

jessicayuen: I'm in favor of the new approach that you have. Something else worth considering is rather than exposing the entire cache via the orchestrator interface, to just add a read only function so that we're not accidentally exposing writes elsewhere. i.e., return a copy of the fetched results, including a copy of the keys so we don't return the original pointers. Thoughts?

lita:
Is it possible to display the output of the admin endpoint here?

LisaLudique:
Added example outputs to the PR description. The config_dump isn't pretty-printed, and I think the solution is to marshal to JSON and pretty-print that but there were comments in the tech spec about making the outputs proto-based?

internal/app/admin/http/handler.go
```diff
@@ -0,0 +1,88 @@
...
+		for _, handler := range handlers {
+			fmt.Fprintf(w, "  %s: %s\n", handler.prefix, handler.description)
+		}
+	}
+}
+
+// TODO(lisalu): Make config output more readable.
+func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		fmt.Fprint(w, bootstrapConfig.String())
```
lita, line 66: Can we json marshell this so that it is formatted? Not sure what the `bootstrapConfig.String()` outputs, but here are some examples of how to use the json Marshall formatter.

https://golangcode.com/json-encode-pretty-print/

internal/app/orchestrator/orchestrator.go
LisaLudique: I agree with having it be read-only and added a "read-only cache" interface that only exposes Fetch.

internal/app/admin/http/handler.go
```diff
@@ -0,0 +1,136 @@
...
+	description string
+	handler     http.HandlerFunc
+}
+
+func getHandlers(bootstrap *bootstrapv1.Bootstrap, orchestrator *orchestrator.Orchestrator) []Handler {
+	handlers := []Handler{
+		{
+			"/",
+			"admin home page",
+			defaultHandler(nil),
```
jessicayuen, line 29: can we omit defining this here if we define it [later](https://github.com/envoyproxy/xds-relay/pull/74/files#diff-29e3fee34912e53bdb13be1567ad0307R43)?

internal/app/admin/http/handler.go
```diff
@@ -0,0 +1,136 @@
...
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, "unable to convert resource to string.\n")
+			return
+		}
+		fmt.Fprint(w, resourceString)
+	}
+}
+
+// Converts an interface to a human-readable string for pretty-printing via marshalling to JSON.
+func interfaceToString(v interface{}) (string, error) {
```
jessicayuen, line 100: May be worth putting this in the util package, for pretty printing debug logs, etc. But we can leave it alone if you think it's unlikely we'd use it outside this package.

internal/app/admin/http/handler.go
```diff
@@ -0,0 +1,136 @@
...
+		configString, err := interfaceToString(bootstrapConfig)
+		if err != nil {
+			fmt.Fprintf(w, "Failed to dump config: %s\n", err.Error())
+		}
+		fmt.Fprintf(w, "%s\n", configString)
+	}
+}
+
+// TODO(lisalu): Support dump of entire cache when no key is provided.
+// TODO(lisalu): Support dump of matching resources when cache key regex is provided.
```
jessicayuen, line 79: It may make sense to add issues for these and follow up in a separate PR.

internal/app/admin/http/handler.go
```diff
@@ -0,0 +1,136 @@
...
+		ExpirationTime: resource.ExpirationTime,
+	}
+
+	return interfaceToString(resourceString)
+}
+
+func getCacheKeyParam(path string) string {
+	// Assumes that the URL is of the format `address/cache/parameter` and returns `parameter`.
+	splitPath := strings.SplitN(path, "/", 3)
+	return splitPath[2]
```
jessicayuen, line 135: Do we need to worry about accessing an out of bounds element here?

internal/app/admin/http/handler_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	handler.ServeHTTP(rr, req)
+	assert.Equal(t, http.StatusOK, rr.Code)
+	assert.Contains(t, rr.Body.String(), `{
+  "Resp": {
+    "Raw": {
+      "version_info": "1",
+      "resources": [
+        {
+          "value": "bGRzIHJlc291cmNl"
```
jessicayuen, line 124: Wonder if there's a way for us to unmarshal this value back to it's original form.

internal/app/cache/cache.go
```diff
@@ -26,6 +26,14 @@ type Cache interface {
...
 	// DeleteRequest removes the given request from any cache entries it's present in.
 	DeleteRequest(key string, req *v2.DiscoveryRequest) error
+
+	// GetReadOnlyCache returns a copy of the cache that only exposes read-only methods in its interface.
+	GetReadOnlyCache() ReadOnlyCache
+}
+
+type ReadOnlyCache interface {
+	// Fetch returns the cached resource if it exists.
+	Fetch(key string) (*Resource, error)
```
jessicayuen, line 36: Should this return `Resource` rather than `*Resource` so we don't accidentally mutate the returned values?

internal/app/mapper/mapper_mock.go
```diff
@@ -0,0 +1,21 @@
+package mapper
```
jessicayuen, line 1: can this live in the mock package too?

internal/app/admin/http/handler.go
LisaLudique: opened #88 and #89 separately from #75

internal/app/mapper/mapper_mock.go
LisaLudique: possibly but this would be a deviation from how mocks in `upstream` are structured

internal/app/admin/http/handler.go
LisaLudique: This shouldn't happen based on how the server does route matching but I'll add a check in case this function is invoked elsewhere or the route matching changes in the future.

internal/app/admin/http/handler.go
LisaLudique: Switched to an empty HandlerFunc. I include it in the list so it shows up in the output for the homepage.

internal/app/admin/http/handler_test.go
```diff
@@ -0,0 +1,178 @@
...
+		Cache:        nil,
+	})
+
+	handler.ServeHTTP(rr, req)
+	assert.Equal(t, http.StatusOK, rr.Code)
+	assert.Equal(t,
+		`{
+  "server": {
+    "address": {
+      "address": "127.0.0.1",
```
eapolinario, line 70: is there any chance the order of the the `server.address` fields are swapped? If so, how about we compare the deserialized json instead?

internal/app/admin/http/handler_test.go
eapolinario: can we use https://godoc.org/github.com/golang/protobuf/jsonpb#Unmarshal ?

internal/app/admin/http/handler_test.go
LisaLudique: can you clarify what you mean by swapped?

eapolinario: Sure. I was asking about the possibility of `port_value` coming before `address`. In case that's possible, that strict comparison would break, right?

LisaLudique: I think it gets unmarshalled based on the order of the fields in the struct, as suggested by https://blog.golang.org/json.

LisaLudique:
I left a TODO and will create an issue for unmarshaling of resources in the cache_dump handler as described in https://envoyproxy.slack.com/archives/CTULFMN91/p1590087624055200?thread_ts=1590087384.054900&cid=CTULFMN91 and will continue that in a separate PR.

internal/app/mapper/mapper_mock.go
eapolinario: What is the idea behind the `mock` package?

LisaLudique: https://github.com/envoyproxy/xds-relay/pull/46#discussion_r410441281

LisaLudique: That said I'm not sure I understand the `*_mock.go` files being separate. Maybe @jyotimahapatra can elaborate?

eapolinario: Isn't the conclusion of that thread that mocks live in the same package as the structures that they are mocking only in a different (called `mock`)?

LisaLudique: `client_mock` doesn't live under `mock` though

eapolinario: I guess what I'm trying to say is that we don't need a package called `mock`. We've been putting mocks under the `mock` as a subdirectory of the structure they are mocking, but under the same package.

internal/app/mapper/mock/mapper.go
```diff
@@ -0,0 +1,11 @@
+package mapper
+
+import (
+	"testing"
+
+	"github.com/envoyproxy/xds-relay/internal/app/mapper"
+)
+
+func NewMapper(t *testing.T) mapper.Mapper {
```
eapolinario, line 9: do you think it'd be cleaner to move this to `internal/app/mapper/mapper_mock.go`? (Same with `/mock/orchestrator.go`)

LisaLudique: I moved all functions from the mock subdirectories (for mapper, orchestartor, and upstream client) to the `*_mock.go` files and deleted the subdirectories. Had to rename some functions but hopefully the structure's less confusing now

internal/app/orchestrator/orchestrator_test.go
```diff
@@ -285,8 +269,8 @@ func TestCachedResponse(t *testing.T) {
 func TestMultipleWatchersAndUpstreams(t *testing.T) {
 	upstreamResponseChannelLDS := make(chan *v2.DiscoveryResponse)
 	upstreamResponseChannelCDS := make(chan *v2.DiscoveryResponse)
+	mapper := mapper.NewMapper(t)
```
jessicayuen, line 272: What do you think of calling these mock instantiators `NewMock`? Reading the test code it's not immediately apparent that they're mocks and how they differ from `mapper.New` unless the package is deliberately named `mock`.

LisaLudique: I agree `NewMock` is clearer

PR #82, opened 2020-05-12
logger: Support formatted logging

jessicayuen:
Equivalent to `fmt.Sprintf`. Support templated logging. This is
especially important when passing the logger to go-control-plane in
certain packages, such as e2e tests.

Without this change we will continue to see log messages like the
following:
```
open watch +%v, %d, ...
```

Also
* Add debug and info logs to orchestrator
* Default to debug logs for e2e tests

### DIFF
--- a/internal/pkg/log/log.go
+++ b/internal/pkg/log/log.go
@@ -33,25 +33,49 @@ type Logger interface {
 	//  ).Info("this is an error message")
 	With(args ...interface{}) Logger
 
-	// Log a message at level Debug, annotated with fields provided through With().
+	// Debug logs a message at level Debug, annotated with fields provided through With().
 	Debug(ctx context.Context, msg ...interface{})
 
-	// Log a message at level Info, annotated with fields provided through With().
+	// Info logs a message at level Info, annotated with fields provided through With().
 	Info(ctx context.Context, msg ...interface{})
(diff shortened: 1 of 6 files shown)


internal/pkg/log/log.go
```diff
@@ -33,26 +33,50 @@ type Logger interface {
...
+	// Debug logs a message at level Debug, annotated with fields provided through With().
 	Debug(ctx context.Context, msg ...interface{})
 
-	// Log a message at level Info, annotated with fields provided through With().
+	// Info logs a message at level Info, annotated with fields provided through With().
 	Info(ctx context.Context, msg ...interface{})
 
-	// Log a message at level Warn, annotated with fields provided through With().
+	// Warn logs a message at level Warn, annotated with fields provided through With().
 	Warn(ctx context.Context, msg ...interface{})
```
lita, line 43: When do we want logging that is not formatted? 🤔 Are these methods still necessary?

jessicayuen: That is a good call out. Since the args are optional, `Warnf` would work where we're using the non-formatted methods currently. I'm going to consolidate these and keep the name of the current functions, `Warn`, `Error` etc. rather than their `<>f` equivalents.

Issue #84, opened 2020-05-12
Integration tests are exhibiting flaky behavior


jessicayuen:
Seems to be failing more frequently:
```
=== RUN   TestXdsClientGetsIncrementalResponsesFromUpstreamServer
2020/05/12 20:31:22 management server listening on 19001
    TestXdsClientGetsIncrementalResponsesFromUpstreamServer: upstream_client_test.go:47: 
        	Error Trace:	upstream_client_test.go:47
        	Error:      	Setup failed: %s
        	Test:       	TestXdsClientGetsIncrementalResponsesFromUpstreamServer
        	Messages:   	rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:19001: connect: connection refused"
```

eapolinario:
As we mentioned during our sync meeting today, these are 2 separate errors. The first one is related to https://github.com/envoyproxy/xds-relay/blob/master/integration/upstream_client_test.go#L208-L218. As @LisaLudique pointed out, a connection refused means that nothing is listening on that port, which can happen if the goroutine that starts the management server doesn't run before we try to connect to it. @jyotimahapatra created an [issue](https://github.com/envoyproxy/xds-relay/blob/master/internal/app/upstream/client.go#L84-L85) to configure the grpc options in that call to the management server, so we could experiment configuring retries there. Fundamentally though, the problem really is because we don't have a way to signal that the management server is ready before we try to connect to it.

As for the second error, we have tests that try to bind to the same management server port before the port is available in the OS. I say this because we [pass in a context](https://github.com/envoyproxy/xds-relay/blob/master/integration/upstream_client_test.go#L208) in the goroutine to start the management server, which [gracefully shuts down the server](https://github.com/envoyproxy/go-control-plane/blob/master/pkg/test/server.go#L85-L87). 

We don't have much insight into how the go runtime schedules the tests, but if I were to guess, at least on Linux we're not leaving enough time for the interaction between the runtime and the OS to  go through the [usual workflow](https://hea-www.harvard.edu/~fine/Tech/addrinuse.html) of closing the TCP connection. In the e2e tests we [wait 1 second](https://github.com/envoyproxy/xds-relay/blob/master/integration/e2e_test.go#L35-L41) between tests, we might adopt a similar approach in the integration tests suite (as a side note,  1s is too much).

eapolinario:
[...]

PR #86, opened 2020-05-14
Start HTTP server for management server in int tests

jessicayuen:
Believe this is the cause of the flakiness, due to shared ports with the
upstream HTTP server in the e2e tests.

### DIFF
--- a/integration/upstream_client_test.go
+++ b/integration/upstream_client_test.go
@@ -27,10 +27,11 @@ import (
 )
 
 const (
-	nodeID           = "node-id"
-	originServerPort = 19001
-	loglevel         = "fatal"
-	updates          = 1
+	nodeID               = "node-id"
+	originServerGRPCPort = 19001
+	originServerHTTPPort = 19002
+	loglevel             = "fatal"
+	updates              = 1
 )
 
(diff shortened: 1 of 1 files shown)


eapolinario:
@jessicayuen , can you clarify why you think that starting the upstream http server fixes the underlying problem? Did you see [my comment](https://github.com/envoyproxy/xds-relay/issues/84#issuecomment-628852414) in the issue? If so, which one of the 2 failures this fix addresses?

integration/upstream_client_test.go
```diff
@@ -205,11 +206,12 @@ func setup(
 	srv3 := gcpserverv3.NewServer(ctx, nil, nil)
 
 	// Start the origin server
-	go gcptest.RunManagementServer(ctx, srv2, srv3, originServerPort)
+	go gcptest.RunHTTP(ctx, originServerHTTPPort)
```
LisaLudique, line 209: What's the distinction between these two servers / why does the origin server need both ports?

jessicayuen:
Going to close this. This is the wrong approach and I think a coincidence that the test is passing. Integration tests shouldn't require the HTTP service because there is no Envoy making the outbound calls.

PR #87, opened 2020-05-18
Add support for stats library

eapolinario:
This PR adds support for emission of metrics using https://github.com/uber-go/tally via the statsd reporter.

A few things to note:
- tally doesn't have support for tags OOB, but if we deem necessary, we can adapt https://github.com/uber/cadence/blob/master/common/metrics/tally/statsd/reporter.go
- I only added support for the statsd as a sink, but extending that to support other reporters is easy
- I tested this locally using https://hub.docker.com/r/hopsoft/graphite-statsd/ and added another script to help debug the emission of stats. We run that container which runs statsd sink and 2 front-end products (grafana and graphite) to aid in development. Instructions on how to visualize the data can be found in that page, but essentially revolves around opening localhost:80 to use grafana or localhost:81/dashboard to use graphite.


internal/app/orchestrator/orchestrator.go
```diff
@@ -122,6 +126,7 @@ func (o *orchestrator) CreateWatch(req gcp.Request) (chan gcp.Response, func())
 	// If this is the first time we're seeing the request from the
 	// downstream client, initialize a channel to feed future responses.
 	responseChannel := o.downstreamResponseMap.createChannel(&req)
+	o.scope.Counter("create_channel").Inc(1)
```
jessicayuen, line 129: Can you make this stats name a const? Ideally we'd also subscope this to `"downstream"` and have the downstreamResponseMap increment this in the `createChannel` function.

eapolinario: good idea.

I added this here just to test, but it makes sense to add it for real, using the convention we're going to use going forward.

eapolinario:
@jessicayuen , re: adding the statsd containter to the e2e, this is a great idea although I'm not sure what sort of verification we can do (i.e. it'd be purely for visual inspection or via introspection in the container). Instead of a makefile rule how about we add a script to ensure that the statsd container is running (and it's exposing the right ports)?

eapolinario:
I pulled the changes to retry integration tests into a separate PR: https://github.com/envoyproxy/xds-relay/pull/91

internal/app/orchestrator/orchestrator.go
```diff
@@ -76,15 +81,18 @@ type orchestrator struct {
...
 	upstreamClient upstream.Client,
 	cacheConfig *bootstrapv1.Cache,
 ) Orchestrator {
 	orchestrator := &orchestrator{
 		logger:                l.Named(component),
+		scope:                 scope,
 		mapper:                mapper,
 		upstreamClient:        upstreamClient,
 		downstreamResponseMap: newDownstreamResponseMap(),
+		downstreamScope:       scope.SubScope("downstream"),
```
jessicayuen, line 95: Ah I meant can you pass this to the above line `newDownstreamResponseMap(scope.Subscope("downstream"))`?

And have the `createChannel` func do the stat increment?

internal/pkg/util/testutils/stats_test.go
```diff
@@ -0,0 +1,62 @@
+package testutils
```
jessicayuen, line 1: Can these tests go in the `stats/` package?

internal/app/orchestrator/orchestrator.go
eapolinario: oh my goodness, obviously. What was I thinking? 😅

internal/app/orchestrator/orchestrator_test.go
```diff
@@ -141,6 +151,9 @@ func TestGoldenPath(t *testing.T) {
 	}
 
 	respChannel, cancelWatch := orchestrator.CreateWatch(req)
+	snap := mockScope.Snapshot()
+	counters := snap.Counters()
+	testutils.AssertCounterValue(t, counters, "mock_orchestrator.downstream.create_channel", 1)
```
LisaLudique, line 156: out of curiosity do we have to worry about flushing stats when testing these?

internal/app/server/server.go
```diff
@@ -63,7 +88,8 @@ func RunWithContext(ctx context.Context, cancel context.CancelFunc, bootstrapCon
 	requestMapper := mapper.NewMapper(aggregationRulesConfig)
 
 	// Initialize orchestrator.
-	orchestrator := orchestrator.New(ctx, logger, requestMapper, upstreamClient, bootstrapConfig.Cache)
+	orchestrator := orchestrator.New(ctx, logger, scope.SubScope("orchestrator"), requestMapper,
```
LisaLudique, line 91: possibly nit, but should scope.SubScope("orchestrator") be declared a const along with metricSubscope?

internal/app/orchestrator/orchestrator_test.go
eapolinario: not for tests since we're using a mock scope. Basically the calls to get a snapshot "do" the job of flushing.

LisaLudique: awesome!

PR #91, opened 2020-05-19
Retry integration tests up to 3 times

eapolinario:
We retry the integration tests up to 3 times using https://github.com/marketplace/actions/retry-step.

This is a stopgap until we find out a better solution to solve issue described in https://github.com/envoyproxy/xds-relay/issues/84#issuecomment-628852414.

I'm adding the failing test just to confirm the retries actually happen.

Signed-off-by: eapolinario <eapolinario@lyft.com>

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -6,8 +6,6 @@ on:
   pull_request:
     branches:
       - master
-env:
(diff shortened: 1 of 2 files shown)


eapolinario:
https://github.com/envoyproxy/xds-relay/pull/91/checks?check_run_id=687289780 is an example of the 3 retries.

PR #94, opened 2020-06-02
Use go-control-plane's PassthroughResponse

jessicayuen:
Context is in this issue: https://github.com/envoyproxy/go-control-plane/issues/299.

### DIFF
--- a/internal/app/cache/cache.go
+++ b/internal/app/cache/cache.go
@@ -8,10 +8,7 @@ import (
 	"time"
 
 	v2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
-	gcp_types "github.com/envoyproxy/go-control-plane/pkg/cache/types"
-	gcp "github.com/envoyproxy/go-control-plane/pkg/cache/v2"
 	"github.com/golang/groupcache/lru"
-	"github.com/golang/protobuf/ptypes/any"
 )
 
 type Cache interface {
@@ -42,14 +39,8 @@ type cache struct {
 	ttl     time.Duration
 }
 
-// Response stores the raw Discovery Response alongside its marshaled resources.
-type Response struct {
-	Raw                v2.DiscoveryResponse
-	MarshaledResources []gcp_types.MarshaledResource
-}
-
 type Resource struct {
-	Resp           *Response
+	Resp           *v2.DiscoveryResponse
 	Requests       map[*v2.DiscoveryRequest]bool
 	ExpirationTime time.Time
 }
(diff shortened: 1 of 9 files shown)


internal/app/orchestrator/orchestrator.go
```diff
@@ -119,7 +119,7 @@ func New(
 //
 // Cancel is an optional function to release resources in the producer. If
 // provided, the consumer may call this function multiple times.
-func (o *orchestrator) CreateWatch(req gcp.Request) (chan gcp.Response, func()) {
+func (o *orchestrator) CreateWatch(req gcp.Request) (chan gcp.ResponseIface, func()) {
```
jessicayuen, line 122: I'm not a huge fan of changing the signature to use `gcp.ResponseIface` everywhere. I may make an upstream PR to change this back to `Response` and rename the struct to something else.

jyotimahapatra: The channel is for an object on which function `GetDiscoveryResponse` have to be called based on an interface. I imagine the usual way in go for representing an interface is using `IFace`?

jessicayuen: I think it may be unintuitive for a couple reasons. The current setup alludes to the [implementation Response](https://github.com/envoyproxy/go-control-plane/blob/b12a9ed045da8cd01b3813e0193d1a381ea9e585/pkg/cache/v2/cache.go#L67-L84) being the defacto implementation, but as seen in places like xds-relay, we don't use it anywhere. I'm also not sure that it's Go convention to explicitly call out that an entity is an interface. It's similar to how the [Resource](https://github.com/envoyproxy/go-control-plane/blob/e933a9f2ccd2c350a7d7c67317fceb79b3cd8ff6/pkg/cache/types/types.go#L8) type in go-control-plane isn't explicitly called `ResourceIface`. I think that's an implementation detail that shouldn't be made transparent for the consumer.

internal/app/cache/cache.go
```diff
@@ -42,14 +39,8 @@ type cache struct {
...
 
-// Response stores the raw Discovery Response alongside its marshaled resources.
-type Response struct {
-	Raw                v2.DiscoveryResponse
-	MarshaledResources []gcp_types.MarshaledResource
-}
-
 type Resource struct {
-	Resp           *Response
+	Resp           *v2.DiscoveryResponse
```
LisaLudique, line 43: this is so much simpler!

LisaLudique:
pending any changes to use Resource rather than ResourceIFace this lgtm

internal/app/orchestrator/orchestrator.go
jyotimahapatra: FWIW. I think `ResponseIFace` does not leak the implementation detail to the consumer.
The goang interface naming convention https://golang.org/doc/effective_go.html#interface-names suggests using a `er` suffix.

eapolinario: Jyoti, IIUC, the naming convention of adding the suffix `er` to an interface only applies to single-method interfaces, which is not the case of the `Response` interface, right?

jyotimahapatra: yeah. Jess's suggestion of sticking with `Response` is better.

PR #97, opened 2020-06-12
Add option to log to filesink

jessicayuen:

### DIFF
--- a/internal/app/orchestrator/orchestrator_test.go
+++ b/internal/app/orchestrator/orchestrator_test.go
@@ -2,7 +2,6 @@ package orchestrator
 
 import (
 	"context"
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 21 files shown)


jessicayuen:
I tested this by starting up the xds-relay server with a bootstrap configuration pointing to a filepath, and saw the logs successfully written there. If we think it's necessary, I can also write an int test for this.

internal/app/orchestrator/orchestrator_test.go
```diff
@@ -55,9 +56,9 @@ func (m mockMultiStreamUpstreamClient) OpenStream(
 
 func newMockOrchestrator(t *testing.T, mockScope tally.Scope, mapper mapper.Mapper,
 	upstreamClient upstream.Client) *orchestrator {
-	logger := log.New("info")
+	logger := log.New("info", os.Stderr)
 	orchestrator := &orchestrator{
-		logger:                logger,
+		logger:                log.New("info", os.Stderr),
```
LisaLudique, line 61: is it worth creating a mocklogger or setting these log levels and stderr as default for testing/no config field provided?

internal/pkg/log/zap.go
```diff
@@ -12,14 +13,18 @@ type logger struct {
 }
 
 // New returns an instance of Logger implemented using the Zap logging framework.
-func New(logLevel string) Logger {
+//
+// logLevel is mandatory, can be one of DEBUG, INFO, WARN, ERROR, FATAL.
+// outputPath is mandatory. Use os.Stderr if unsure.
```
LisaLudique, line 18: nit: writeTo

internal/app/orchestrator/orchestrator_test.go
jessicayuen: that's a good idea, I can do it as a part of this PR.

Issue #93, opened 2020-05-26
Resource unmarshalling in cache_dump admin API


LisaLudique:
The resources in the DiscoveryResponse are still in their marshalled any.Any form when fetched with the cache_dump API (https://github.com/envoyproxy/xds-relay/pull/74/files#r426102814). The cache_dump API should inspect the type_URL to unmarshal the Any variable into the correct resource to make it more readable: https://envoyproxy.slack.com/archives/CTULFMN91/p1590087624055200?thread_ts=1590087384.054900&cid=CTULFMN91.

PR #99, opened 2020-06-16
Unmarshal v2 resources in cache_dump

LisaLudique:
This change unmarshals xDS resources in the cache from the generic Any form based on the TypeURL, utilizing the same logic as https://github.com/envoyproxy/go-control-plane/blob/e933a9f2ccd2c350a7d7c67317fceb79b3cd8ff6/pkg/cache/v2/resource.go#L52. This change only unmarshals V2 resources. Resources whose type URLs do not match are included in the output under the `Unmarshalled` field. Addresses #93.

Example output:
```
{
  "Resp": {
    "VersionInfo": "1",
    "Resources": {
      "Endpoints": null,
      "Clusters": null,
      "Routes": null,
      "Listeners": [
        {
          "name": "lds resource"
        }
      ],
      "Secrets": null,
      "Runtimes": null,
      "Unmarshalled": null
    },
    "Canary": false,
    "TypeURL": "type.googleapis.com/envoy.api.v2.Listener",
    "Nonce": "",
    "ControlPlane": null
  },
  "Requests": [
    {
      "type_url": "type.googleapis.com/envoy.api.v2.Listener"
    }
  ],
  "ExpirationTime": "2020-06-16T10:45:46.703325-07:00"
}
```

### DIFF
--- a/internal/app/admin/http/handler.go
+++ b/internal/app/admin/http/handler.go
@@ -143,12 +143,13 @@ type marshalledDiscoveryResponse struct {
 }
 
 type xDSResources struct {
-	Endpoints []*v2.ClusterLoadAssignment
-	Clusters  []*v2.Cluster
-	Routes    []*v2.RouteConfiguration
-	Listeners []*v2.Listener
-	Secrets   []*envoy_api_v2_auth.Secret
-	Runtimes  []*envoy_service_discovery_v2.Runtime
+	Endpoints    []*v2.ClusterLoadAssignment
+	Clusters     []*v2.Cluster
+	Routes       []*v2.RouteConfiguration
+	Listeners    []*v2.Listener
+	Secrets      []*envoy_api_v2_auth.Secret
+	Runtimes     []*envoy_service_discovery_v2.Runtime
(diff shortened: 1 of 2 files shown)


internal/app/admin/http/handler.go
```diff
@@ -102,30 +110,108 @@ func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
...
+			if err == nil {
+				marshalledResources.Secrets = append(marshalledResources.Secrets, s)
+			}
+		case resource2.RuntimeType:
+			r := &envoy_service_discovery_v2.Runtime{}
+			err := ptypes.UnmarshalAny(resource, r)
+			if err == nil {
+				marshalledResources.Runtimes = append(marshalledResources.Runtimes, r)
+			}
+		default:
```
jessicayuen, line 209: Do you think we should add a catch all `[]*any.Any` in the `xDSResources` struct in this scenario and include that in the output?

LisaLudique: Added

PR #100, opened 2020-06-17
Support admin endpoint for updating the logging level

LisaLudique:
This change adds support for an admin endpoint that updates the logging level, via internally instantiating a new logger at the specified level. Currently it does not restrict the possible zap logging levels the logger can be set to.

It's worth noting that updates to the logging level are not reflected in the `/server_info` endpoint, which only reflects the bootstrap config that the server is started with (unlike envoy's `/config_dump` which reflects the current state).


internal/app/admin/http/handler.go
```diff
@@ -34,6 +39,11 @@ func getHandlers(bootstrap *bootstrapv1.Bootstrap, orchestrator *orchestrator.Or
 			"print cache entry for a given key. usage: `/cache/<key>`",
 			cacheDumpHandler(orchestrator),
 		},
+		{
+			"/log_level/",
+			"update the log level. usage: `/log_level/<level>`",
```
jessicayuen, line 44: May want to consider listing the options here: debug, info, warn, error, fatal

internal/app/admin/http/handler.go
```diff
@@ -126,11 +138,29 @@ func resourceToString(resource cache.Resource) (string, error) {
...
 		return splitPath[2], nil
 	}
-	return "", fmt.Errorf("unable to parse cache key from path: %s", path)
+	return "", fmt.Errorf("unable to parse parameter from path: %s", path)
+}
+
+func logLevelHandler(l log.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		if req.Method == "POST" {
+			logLevel, _ := getParam(req.URL.Path)
```
jessicayuen, line 153: If this returns an error I think we should exit. `zap.ParseLogLevel` will default to Info IIRC.  May produce unexpected behavior.

internal/app/admin/http/handler.go
LisaLudique: Is it worth restricting the possible levels to be consistent with bootstrap.proto? What's the reasoning for restricting the possible levels?

internal/app/admin/http/handler.go
LisaLudique: I think that's what it currently does? Returns a BadRequest error to the client

jessicayuen: we don't assign the error in this line though. If we can't parse the param in `getParam`, `zap.ParseLogLevel("")` is effectively called, which defaults to the info log level, no?

LisaLudique: oh, I was looking at the wrong line. Yeah we should catch that error I don't know why I didn't include it.

jessicayuen: actually nvm, I think this is me misreading the implementation for `zap.ParseLogLevel`.

internal/app/admin/http/handler.go
jessicayuen: I just meant listing the levels specified in the proto here to help guide the endpoint user easier (without having to search up the proto spec).

jessicayuen:
lgtm, will +1 once you resolve the conflicts!

internal/app/admin/http/handler.go
```diff
@@ -126,11 +138,33 @@ func resourceToString(resource cache.Resource) (string, error) {
...
+	return "", fmt.Errorf("unable to parse parameter from path: %s", path)
+}
+
+func logLevelHandler(l log.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		if req.Method == "POST" {
+			logLevel, err := getParam(req.URL.Path)
+			if err != nil {
+				w.WriteHeader(http.StatusBadRequest)
+				fmt.Fprintf(w, "unable to parse log level from path: %s", err.Error())
```
jessicayuen, line 156: missing `return`?

PR #101, opened 2020-06-24
Full cache dump support for admin API

LisaLudique:
Adds support for outputting the entirety of the cache via admin API.

### DIFF
--- a/internal/app/orchestrator/downstream.go
+++ b/internal/app/orchestrator/downstream.go
@@ -14,6 +14,8 @@ package orchestrator
 import (
 	"sync"
 
+	"github.com/envoyproxy/xds-relay/internal/app/mapper"
+
 	gcp "github.com/envoyproxy/go-control-plane/pkg/cache/v2"
 	"github.com/uber-go/tally"
 )
@@ -85,4 +87,19 @@ func (d *downstreamResponseMap) deleteAll(watchers map[*gcp.Request]bool) {
 			delete(d.responseChannels, watch)
 		}
 	}
+}
+
+func (d *downstreamResponseMap) keys(m *mapper.Mapper) []string {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	var keys []string
+	for request := range d.responseChannels {
+		key, err := mapper.Mapper.GetKey(*m, *request)
+		if err != nil {
+			// TODO(lisalu): Log warning.
+		} else {
+			keys = append(keys, key)
+		}
+	}
+	return keys
 }
--- a/internal/app/orchestrator/downstream.go
+++ b/internal/app/orchestrator/downstream.go
@@ -81,17 +81,16 @@ func (d *downstreamResponseMap) deleteAll(watchers map[*gcp.Request]bool) {
 	}
 }
 
-func (d *downstreamResponseMap) keys(m *mapper.Mapper) []string {
+func (d *downstreamResponseMap) keys(m *mapper.Mapper) ([]string, error) {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
 	var keys []string
 	for request := range d.responseChannels {
 		key, err := mapper.Mapper.GetKey(*m, *request)
 		if err != nil {
-			// TODO(lisalu): Log warning.
-		} else {
-			keys = append(keys, key)
+			return nil, err
 		}
+		keys = append(keys, key)
 	}
-	return keys
+	return keys, nil
 }
--- a/internal/app/orchestrator/downstream.go
+++ b/internal/app/orchestrator/downstream.go
@@ -81,16 +81,18 @@ func (d *downstreamResponseMap) deleteAll(watchers map[*gcp.Request]bool) {
 	}
 }
 
-func (d *downstreamResponseMap) keys(m *mapper.Mapper) ([]string, error) {
+// getAggregatedKeys returns a list of aggregated keys for all requests in the downstream response map.
+func (d *downstreamResponseMap) getAggregatedKeys(m *mapper.Mapper) (map[string]bool, error) {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
-	var keys []string
+	// Since multiple requests can map to the same cache key, we use a map to ensure unique entries.
+	keys := make(map[string]bool)
 	for request := range d.responseChannels {
 		key, err := mapper.Mapper.GetKey(*m, *request)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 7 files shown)


LisaLudique:
cc @jessicayuen @eapolinario for initial thoughts on this approach. It's not too big of a change. The main downside is that the keys from the downstream response map and what's in the cache may not be in sync by the time we call Fetch with them.

internal/app/orchestrator/orchestrator.go
```diff
@@ -210,6 +212,10 @@ func (o *orchestrator) GetReadOnlyCache() cache.ReadOnlyCache {
 	return o.cache.GetReadOnlyCache()
 }
 
+func (o *orchestrator) GetKeys() []string {
```
jessicayuen, line 215: What do you think of combining this with `GetReadOnlyCache()` to return `(cache.ReadOnlyCache, []string)`? I'd rather expose less interface functions from the Orchestrator if these two functions are meant to be used in sequence.

jessicayuen: This way you can also put a lock around these operations so we don't have the sync issue you mentioned above.

LisaLudique: Can you clarify this approach? I think there would be an issue with this because one comes from the cache and one comes from the downstream response map, so there isn't a lock that would encompass operations on both.

LisaLudique:
cleaned up the tests. I'm going to prettify the output a bit, add logs, and move the mock upstream clients into the client_mock file.

LisaLudique:
```=== RUN   TestGoldenPath
137
{"level":"debug","ts":1594079119.200589,"msg":"creating watch","node ID":"","type":"type.googleapis.com/envoy.api.v2.Listener","version":"","nonce":"","error detail":"<nil>","aggregated key":"lds"}
138
{"level":"debug","ts":1594079119.2006137,"logger":"cache","msg":"request added","key":"lds","node ID":"","request type":"type.googleapis.com/envoy.api.v2.Listener"}
139
{"level":"debug","ts":1594079119.2006288,"logger":"cache","msg":"fetch","key":"lds","version":"","type_url":"","resource length":0}
140
{"level":"debug","ts":1594079119.200641,"msg":"watching upstream","key":"lds"}
141
{"level":"debug","ts":1594079119.2007303,"logger":"cache","msg":"set response","key":"lds","response url":"type.googleapis.com/envoy.api.v2.Listener"}
142
{"level":"debug","ts":1594079119.2007456,"logger":"cache","msg":"fetch","key":"lds","version":"1","type_url":"type.googleapis.com/envoy.api.v2.Listener","resource length":1}
143
{"level":"debug","ts":1594079119.2007554,"msg":"response fanout initiated","key":"lds","response type":"type.googleapis.com/envoy.api.v2.Listener","response version":"1"}
144
--- FAIL: TestGoldenPath (0.00s)
145
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
146
	panic: runtime error: invalid memory address or nil pointer dereference
147
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xa4d7c9]
148

149
goroutine 23 [running]:
150
testing.tRunner.func1.1(0xb1d440, 0x134c5f0)
151
	/opt/hostedtoolcache/go/1.14.4/x64/src/testing/testing.go:940 +0x2f5
152
testing.tRunner.func1(0xc0004db560)
153
	/opt/hostedtoolcache/go/1.14.4/x64/src/testing/testing.go:943 +0x3f9
154
panic(0xb1d440, 0x134c5f0)
155
	/opt/hostedtoolcache/go/1.14.4/x64/src/runtime/panic.go:969 +0x166
156
github.com/envoyproxy/xds-relay/internal/pkg/util/testutils.AssertCounterValue(0xc0004db560, 0xc0000a8000, 0xc000120240, 0x22, 0x1)
157
	/home/runner/work/xds-relay/xds-relay/go/src/github.com/envoyproxy/xds-relay/internal/pkg/util/testutils/stats.go:14 +0xa9
158
github.com/envoyproxy/xds-relay/internal/app/orchestrator.TestGoldenPath(0xc0004db560)
159
	/home/runner/work/xds-relay/xds-relay/go/src/github.com/envoyproxy/xds-relay/internal/app/orchestrator/orchestrator_test.go:159 +0x88e
160
testing.tRunner(0xc0004db560, 0xc70f60)
161
	/opt/hostedtoolcache/go/1.14.4/x64/src/testing/testing.go:991 +0xdc
162
created by testing.(*T).Run
163
	/opt/hostedtoolcache/go/1.14.4/x64/src/testing/testing.go:1042 +0x357
164
FAIL	github.com/envoyproxy/xds-relay/internal/app/orchestrator	0.030s
```

internal/app/orchestrator/downstream.go
```diff
@@ -78,3 +80,17 @@ func (d *downstreamResponseMap) deleteAll(watchers map[*gcp.Request]bool) {
 		}
 	}
 }
+
+func (d *downstreamResponseMap) keys(m *mapper.Mapper) ([]string, error) {
```
jessicayuen, line 84: nit: maybe change this to `requestKeys` to not conflate with the keys for the map which are of `Request` type.

internal/app/orchestrator/downstream.go
```diff
@@ -78,3 +80,17 @@ func (d *downstreamResponseMap) deleteAll(watchers map[*gcp.Request]bool) {
...
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	var keys []string
```
jessicayuen, line 87: Do we need to dedupe since 2 requests can map to the same keys?

internal/app/orchestrator/orchestrator.go
```diff
@@ -60,6 +60,8 @@ type Orchestrator interface {
 	shutdown(ctx context.Context)
 
 	GetReadOnlyCache() cache.ReadOnlyCache
+
+	GetKeys() ([]string, error)
```
jessicayuen, line 64: same, `GetDownstreamAggregatedKeys()`?

internal/app/orchestrator/orchestrator.go
jessicayuen: Sorry I just saw this comment. The aggregated key should usually be 1:1 with the cache keys, but I see your point. Let's leave it as it. 👍

internal/app/orchestrator/downstream.go
LisaLudique: yep, good catch.

PR #102, opened 2020-06-25
Add stats to main workflows

jessicayuen:
Add stats to:
* orchestrator
* mapper
* cache
* upstream


jessicayuen:
I added the upstream stats. This is a preliminary list, I think we can add more as needed. Let me know your thoughts!

PR #109, opened 2020-07-20
Add Error Interceptor

eapolinario:
This PR adds a [`StreamClientInterceptor`](https://github.com/grpc/grpc-go/tree/master/examples/features/interceptor#stream-interceptor) that records whenever there's an error in either `SendMsg` or `RecvMsg` functions in any of the calls in the bidi streams present in the upstream client.

Signed-off-by: eapolinario <eapolinario@lyft.com>

### DIFF
--- a/internal/app/metrics/metrics.go
+++ b/internal/app/metrics/metrics.go
@@ -89,9 +89,9 @@ const (
 	MapperError   = "error"   // counter, # of errors when converting a request to an aggregated key
 )
 
-// .error_interceptor
+// .upstream.error_interceptor
 const (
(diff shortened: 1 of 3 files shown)


internal/app/metrics/metrics.go
```diff
@@ -89,6 +89,15 @@ const (
 	MapperError   = "error"   // counter, # of errors when converting a request to an aggregated key
 )
 
+// .error_interceptor
+const (
+	// scope: .error_interceptor.*
```
jessicayuen, line 94:
```suggestion
	// scope: .upstream.error_interceptor.*
```

PR #110, opened 2020-07-21
Document aggregation keyer rules

jessicayuen:

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,3 +11,157 @@ Caching, aggregation, and relaying for xDS compliant clients and origin servers
 To get started:
 
 * [Contributing guide](CONTRIBUTING.md)
+
+## Design
+
+## Aggregation Rules
+
+xds-relay uses aggregation rules to determine how to aggregate inbound connections to the origin
+management server ([example](https://github.com/envoyproxy/xds-relay/blob/c83a956952c3e9762a08420ffc1c6bda5e0fff04/integration/testdata/keyer_configuration_e2e.yaml)
+).
+
+xDS requests that match under the same set of aggregation rules will fall into the same cache
+bucket. The cache is a mapping of a aggregation key to a xDS [DiscoveryResponse](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/discovery.proto#discoveryresponse)
+. This implies that all inbound xDS requests that map to the same aggregation key will retrieve the
+same DiscoveryResponse. This is useful in multi-cluster scenarios where xDS configuration does not
+differ greatly between clusters. For example, a service application scaled across 100 hosts will
+likely require the same data plane configurations across all hosts.
+
+At a high level, an aggregation rule is formed from `fragments`, `rules`, `match`, and `result`s.
+
+#### Fragments
+A final aggregation key may look like `fooservice_production_lds`. Each of those identifiers
+separated by `_` is what we call a fragment. _fooservice_, _production_, and _lds_ are 3
+fragments in the previous example.
+
+#### Rules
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,3 +11,157 @@ Caching, aggregation, and relaying for xDS compliant clients and origin servers
...
+              - "type.googleapis.com/envoy.api.v2.Cluster"
+        result:
+          request_node_fragment:
+            field: 1
+            action:
+              regex_action: { pattern: "^*-(.*)-*$", replace: "$1" }
+```
+This example specifies that if the DiscoveryRequest type matches LDS or CDS, to generate the first
+fragment from the DiscoveryRequest node ID field by performing a regex replace operation using the
+pattern: `"^(.*)-*$"`, and replacing it with the first regex group match `"$1"`. More concretely,
```
LisaLudique, line 61: nit: should be "^*-(.*)-*$"

LisaLudique: can you also define `field` here? does that correspond to "$1"?

README.md
```diff
@@ -11,3 +11,157 @@ Caching, aggregation, and relaying for xDS compliant clients and origin servers
...
+
+#### Fragments
+A final aggregation key may look like `fooservice_production_lds`. Each of those identifiers
+separated by `_` is what we call a fragment. _fooservice_, _production_, and _lds_ are 3
+fragments in the previous example.
+
+#### Rules
+Each `fragment` can have multiple `rule`s ([example](https://github.com/envoyproxy/xds-relay/blob/c83a956952c3e9762a08420ffc1c6bda5e0fff04/integration/testdata/keyer_configuration_e2e.yaml#L28-L52)
+). A rule defines how to generate a fragment from an xDS [DiscoveryRequest](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/discovery.proto#discoveryrequest)
+. The first matching rule applies, and the rest are ignored.
```
eapolinario, line 40: I don't think this is true, i.e. we can have multiple matches in a list of rules. The mapper currently [matches](https://github.com/envoyproxy/xds-relay/blob/c83a956952c3e9762a08420ffc1c6bda5e0fff04/internal/app/mapper/mapper.go#L49-L66) fragments for all rules, not only the first one (as [this comment](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto#L27-L28) implies).

README.md
```diff
@@ -11,3 +11,157 @@ Caching, aggregation, and relaying for xDS compliant clients and origin servers
...
+).
+
+xDS requests that match under the same set of aggregation rules will fall into the same cache
+bucket. The cache is a mapping of a aggregation key to a xDS [DiscoveryResponse](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/discovery.proto#discoveryresponse)
+. This implies that all inbound xDS requests that map to the same aggregation key will retrieve the
+same DiscoveryResponse. This is useful in multi-cluster scenarios where xDS configuration does not
+differ greatly between clusters. For example, a service application scaled across 100 hosts will
+likely require the same data plane configurations across all hosts.
+
+At a high level, an aggregation rule is formed from `fragments`, `rules`, `match`, and `result`s.
```
eapolinario, line 30: We should give a better description of how each one of these 4 components interact. For example, aggregations rules are composed of `fragments`, and a single `fragment` is composed of a list of `rules` where and each rule is composed of a pair composed of a `match` predicate and a `result` predicate.

Do you think it's too much to try to specify this using [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form)? It might be, I'm just wondering if having that could alleviate the problem of being total in this document.

PR #115, opened 2020-07-28
Explicitly close downstream watches

jessicayuen:
Although the language specification for Go specifies that channels will
be garbage collected when no longer referenced, the receiver (
go-control-plane in this scenario) may be still looking for an explicit
signal. In our experiments, we saw that there were scenarios where cache
TTL triggering removal of downstream watchers, did not cause the Envoy
clients to resend requests.

### DIFF
--- a/internal/app/metrics/metrics.go
+++ b/internal/app/metrics/metrics.go
@@ -27,7 +27,7 @@ const (
 	// scope: .orchestrator.$aggregated_key.cache_evict.*
 	ScopeOrchestratorCacheEvict            = "cache_evict"
 	OrcheestratorCacheEvictCount           = "calls"            // counter, # of times cache evict is called
-	OrchestratorOnCacheEvictedRequestCount = "requests_evicted" // counter, # of requests that were evicted for this aggregated key.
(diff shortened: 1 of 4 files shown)


jessicayuen:
Primary concern here is a race between the two operations (get response channel, add response) here:
https://github.com/envoyproxy/xds-relay/blob/37af3c5c5e65772d6bb5b0841ca117ff09c99c60/internal/app/orchestrator/orchestrator.go#L340-L341

We can make this an atomic operation and add a lock but then we're blocking fanouts of other routines.

eapolinario:
@jessicayuen , this looks good (minus the issue you already mentioned above). I have only one question: can we double check if closing the channel handed by `go-control-plane` is enough to close the connection?

jessicayuen:
> @jessicayuen , this looks good (minus the issue you already mentioned above). I have only one question: can we double check if closing the channel handed by go-control-plane is enough to close the connection?

To close the loop here, we discussed offline and verified from Eduardo's sandbox setup that closing the channel indeed forces the Envoy client to re-issue a new request.

PR #129, opened 2020-08-21
Remove redirect behavior for admin endpoints and support getting log level

### LINKED ISSUE #127: Have admin endpoint /cache behave the same as /cache/
With Go http handler behavior, using just `/cache` results in a redirect as the handler tries to send it to `/cache/`.

LisaLudique:
This change:

- Supports returning the current log level in string format when no parameter is supplied to the admin logging endpoint.
- Handles trailing slashes in URLs, such that paths with and without are registered to the same handlers without any redirects. [The redirects are a characteristic of http handlers in Go](https://golang.org/pkg/net/http/#ServeMux):
```If a subtree has been registered and a request is received naming the subtree root without its trailing slash, ServeMux redirects that request to the subtree root (adding the trailing slash). This behavior can be overridden with a separate registration for the path without the trailing slash. For example, registering "/images/" causes ServeMux to redirect a request for "/images" to "/images/", unless "/images" has been registered separately.```

Fixes #127

### DIFF
--- a/internal/app/admin/http/handler.go
+++ b/internal/app/admin/http/handler.go
@@ -27,7 +27,7 @@ import (
 )
 
 type Handler struct {
-	prefix      string
+	pattern     string
 	description string
 	handler     http.HandlerFunc
 }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: internal/app/admin/http/handler.go [hunk 1]
```diff
@@ -261,24 +260,27 @@ func marshalResources(Resources []*any.Any) *xDSResources {
...
 		if req.Method == "POST" {
-			logLevel, err := getParam(req.URL.Path)
-			if err != nil {
-				w.WriteHeader(http.StatusBadRequest)
-				fmt.Fprintf(w, "unable to parse log level from path: %s", err.Error())
+			logLevel := getParam(req.URL.Path)
+
+			// If no key is provided, output the current log level.
+			if logLevel == "" {
+				fmt.Fprintf(w, "Current log level: %s\n", l.GetLevel())
```
jessicayuen, line 279: Thoughts on also outputting this message on success scenarios?

internal/pkg/log/zap.go
```diff
@@ -58,6 +59,12 @@ func (l *logger) UpdateLogLevel(logLevel string) {
 
 	log = log.With(z.Namespace("json"))
 	l.zap = log.Sugar()
+	l.level = zLevel.String()
+}
+
+// GetLevel returns the logging level in human-readable string format..
```
jessicayuen, line 65: nit: extra `.`

PR #131, opened 2020-08-24
Add Request/Response abstraction

jyotimahapatra:

As an initial step to introduce v3 api, we need to separate the logic of orchestrator such that there are no dependencies on v2/v3 DiscoveryRequest.
This PR removes the dependency on the response channels and creates a watch interface to send responses.

Next PR: abstract DiscoveryRequest from orchestrator and mapper.

### DIFF
--- a/internal/app/orchestrator/downstream.go
+++ b/internal/app/orchestrator/downstream.go
@@ -25,8 +25,8 @@ import (
 // WaitGroup is used here to ensure that all writes to the channel are
 // processed before we attempt to close the channel.
 type responseChannel struct {
-	wg sync.WaitGroup
-	w  transport.Watch
+	wg    sync.WaitGroup
+	watch transport.Watch
 }
 
 // downstreamResponseMap is a map of downstream xDS client requests to response
@@ -49,7 +49,7 @@ func (d *downstreamResponseMap) createChannel(req *gcp.Request) *responseChannel
 	defer d.mu.Unlock()
 	if _, ok := d.responseChannels[req]; !ok {
 		d.responseChannels[req] = &responseChannel{
-			w: transport.NewWatchV2(req),
+			watch: transport.NewWatchV2(req),
 		}
 	}
 	return d.responseChannels[req]
@@ -71,9 +71,9 @@ func (d *downstreamResponseMap) delete(req *gcp.Request) transport.Watch {
 	if responseChannel, ok := d.responseChannels[req]; ok {
 		// wait for all writes to the responseChannel to complete before closing.
 		responseChannel.wg.Wait()
-		responseChannel.w.Close()
+		responseChannel.watch.Close()
 		delete(d.responseChannels, req)
-		return responseChannel.w
+		return responseChannel.watch
 	}
 	return nil
 }
@@ -87,7 +87,7 @@ func (d *downstreamResponseMap) deleteAll(watchers map[*gcp.Request]bool) {
 		if responseChannel, ok := d.responseChannels[watch]; ok {
 			// wait for all writes to the responseChannel to complete before closing.
 			responseChannel.wg.Wait()
-			responseChannel.w.Close()
+			responseChannel.watch.Close()
 			delete(d.responseChannels, watch)
 		}
 	}
@@ -112,7 +112,7 @@ func (d *downstreamResponseMap) getAggregatedKeys(m *mapper.Mapper) (map[string]
 func (responseChannel *responseChannel) addResponse(resp gcp.PassthroughResponse) error {
 	responseChannel.wg.Add(1)
 	defer responseChannel.wg.Done()
-	ok, err := responseChannel.w.Send(resp)
+	ok, err := responseChannel.watch.Send(resp)
 	if err != nil {
 		return err
 	}
--- a/internal/app/orchestrator/downstream.go
+++ b/internal/app/orchestrator/downstream.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 
 	"github.com/envoyproxy/xds-relay/internal/app/mapper"
+	"github.com/envoyproxy/xds-relay/internal/app/transport"
 
 	gcp "github.com/envoyproxy/go-control-plane/pkg/cache/v2"
 )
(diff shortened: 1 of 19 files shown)


internal/app/orchestrator/downstream.go
```diff
@@ -24,8 +25,8 @@ import (
 // WaitGroup is used here to ensure that all writes to the channel are
 // processed before we attempt to close the channel.
 type responseChannel struct {
-	wg      sync.WaitGroup
-	channel chan gcp.Response
+	wg sync.WaitGroup
+	w  transport.Watch
```
jessicayuen, line 29: can we explicitly call this `watch` or `transport`?

internal/app/transport/watch.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+
+	gcp "github.com/envoyproxy/go-control-plane/pkg/cache/v2"
+)
+
+// Watch interface abstracts v2 and v3 watches
+type Watch interface {
+	Close()
+	GetCh() interface{}
```
jessicayuen, line 12: Hmm I prefer to avoid the type conversion from `interface{}` where we can. Anything we can do to encapsulate this since this is either of type `chan v2.Response` or type `chan v3.Response`?

internal/app/transport/watch.go
```diff
@@ -0,0 +1,54 @@
...
+	Send(s interface{}) (bool, error)
+}
+
+var _ Watch = &WatchV2{}
+
+// WatchV2 is the transport object that takes care of send responses to the xds clients
+type WatchV2 struct {
+	Req *gcp.Request
```
jessicayuen, line 20: does this var need to be exported?

jyotimahapatra: I had to export due to a linter [error](https://github.com/golang/lint/issues/465) previously while working in EM.

internal/app/transport/watch.go
jyotimahapatra: we could take either approach. We could expose 2 methods instead with concrete signature if you prefer. Let me do that.

internal/app/orchestrator/downstream.go
jyotimahapatra: its in a separate `transport` package. Are you suggesting keeping the watch in the orchestrator package?

internal/app/transport/watch.go
jyotimahapatra: looks like linter does not complain here. making watchv2 unexported. Thanks for raising this.

internal/app/transport/watch.go
jyotimahapatra: let me know if the two methods look better now.

internal/app/orchestrator/downstream.go
jessicayuen: ah I just mean changing the variable name `w`.

jyotimahapatra: ah right...doing that now

jyotimahapatra: done

internal/app/transport/watch.go
jessicayuen: Would something like this work? I think with the two methods the interface becomes less valuable because it starts to get an understanding of the v2/v3 implementation.

```
func (watch *watchV2) GetChan() ResponseCh { return Response{V2: watch.out} }
func (watch *watchV3) GetChan() ResponseCh { return Response{V3: watch.out} }

type ResponseCh struct {
     V2 chan gcpV2.Response
     V3 chan gcpV3.Response
}

type Watch interface {
    Close()
    GetCh() ResponseCh
    ...
}
```

From the caller:
```
w.GetCh().V2
```

Would suggest doing something similar with `Send(interface{})` if it works.

jyotimahapatra: It won't be possible with `Send`, because it is going to be called from version agnostic code in orchestrator. You are right about the inefficacy of the two methods, and that was the purpose of using interface earlier. We are going to add the Request/Response abstractions in the next PR and could show how the interfaces are going to interact. The watch would be called only from the transport package only.

jyotimahapatra: If it helps, I can put out a bigger PR about how Request/Response abstractions are going to come through in the subsequent PRs and how the usage of interface will be limited to transport package only, and will be unit tested.

jyotimahapatra: I've made changes to include all interfaces, Request/Response/Watch to see the interfaces better. We can iterate on these changes now. Hopefully the changes aren't too big.

internal/app/orchestrator/orchestrator.go
```diff
@@ -129,48 +130,54 @@ func New(
...
 	if err != nil {
 		// If we fail to register the watch, we need to kill this stream by
 		// closing the response channel.
 		o.logger.With("error", err).With("aggregated_key", aggregatedKey).With(
-			"request_node", req.GetNode()).Error(ctx, "failed to add watch")
+			"request", req.GetRaw().V2).Error(ctx, "failed to add watch")
 		metrics.OrchestratorWatchErrorsSubscope(o.scope, aggregatedKey).Counter(metrics.ErrorRegisterWatch).Inc(1)
-		closedChannel := o.downstreamResponseMap.delete(&req)
-		return closedChannel, nil
+		w := o.downstreamResponseMap.delete(req)
```
jessicayuen, line 179: nit: name this `closedWatch`

internal/app/transport/request.go
```diff
@@ -0,0 +1,123 @@
...
+	status "google.golang.org/genproto/googleapis/rpc/status"
+)
+
+// RequestVersion holds either one of the v2/v3 DiscoveryRequests
+type RequestVersion struct {
+	V2 *discoveryv2.DiscoveryRequest
+}
+
+// Request is the generic interface to abstract v2 and v3 DiscoveryRequest types
+type Request interface {
```
jessicayuen, line 15: I wonder if it makes sense to move this to go-control-plane. Doesn't have to be done now.

jyotimahapatra: Sure, we could try to remove the duplicated code in go-control-plane with something like this and move it there subsequently.

jyotimahapatra:
@jessicayuen I had to push some more changes to fix a few last compilation errors. PTAL again.

PR #133, opened 2020-08-25
Rename Dockerfile to Dockerfile-e2e-tests

nezdolik:
As current Dockerfile in repo is used purely for e2e tests, it could be misleading for developers. Lets rename it to highlight the purpose better.


.github/workflows/end2end-tests.yml
```diff
@@ -18,6 +18,6 @@ jobs:
           path: ${{ env.PACKAGEPATH }}
       - name: Build image
         working-directory: ${{ env.PACKAGEPATH }}
-        run: docker build . --file Dockerfile --tag xdsrelay
+        run: docker build . --file Dockerfile-e2e-tests --tag xdsrelay
```
eapolinario, line 21: does running `make build-docker-image` work here?

Makefile
```diff
@@ -34,7 +34,7 @@ compile-validator-tool: setup  ## Compiles configuration validator tool
 
 .PHONY: build-docker-image
 build-docker-image: ## Build docker image for use in e2e tests
-	docker build . --file Dockerfile --tag xds-relay
```
eapolinario, line 37: how about renaming this make rule to `build-e2e-tests-docker-image`?

.github/workflows/end2end-tests.yml
nezdolik: it does (as`--file` argument allows to provide custom file name)

Makefile
nezdolik: fixed

.github/workflows/end2end-tests.yml
eapolinario: In that case how about we just run `make build-e2e-tests-docker-image` here?

nezdolik: cannot yet explain the reason, but suggested change makes the next step "run e2e tests inside container" fail with error:
`Run docker run -v /home/runner/work/xds-relay/xds-relay/go/src/github.com/envoyproxy/xds-relay:/xds-relay xdsrelay:latest make e2e-tests
Unable to find image 'xdsrelay:latest' locally
docker: Error response from daemon: pull access denied for xdsrelay, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.`

eapolinario: interesting. Github Actions is probably detects the use of `docker build` in the command and sets up the right permissions. That said, I couldn't find anything in the docs though. Feel free to revert https://github.com/envoyproxy/xds-relay/pull/133/commits/48087da1a2280021dbfa0d4fbd6f846fd50d3e73.

nezdolik:
Reverted last commit, looks like there is a bug with e2e test:
```    e2e_test.go:106: 
        Timed out after 1.000s.
        Error: Unexpected non-nil/non-zero extra argument at index 1:
        	<int>: 9```

PR #135, opened 2020-08-28
Abstract some more remaining use of v2 protos

jyotimahapatra:
Removing some last few references to V2 in orchestrator code

### DIFF
--- a/internal/app/transport/request.go
+++ b/internal/app/transport/request.go
@@ -6,8 +6,8 @@ import (
 	status "google.golang.org/genproto/googleapis/rpc/status"
 )
 
-// requestVersion holds either one of the v2/v3 DiscoveryRequests
-type requestVersion struct {
+// RequestVersion holds either one of the v2/v3 DiscoveryRequests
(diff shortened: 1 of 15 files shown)


internal/app/transport/request.go
```diff
@@ -107,9 +111,14 @@ func (r *RequestV2) GetSubZone() string {
...
-// GetRaw gets the error details
-func (r *RequestV2) GetRaw() *RequestVersion {
-	return &RequestVersion{V2: r.r}
+// GetVersionedRequest gets the error details
+func (r *RequestV2) getVersionedRequest() *requestVersion {
+	return &requestVersion{V2: r.r}
+}
+
+// GetRequest gets the error details
+func (r *RequestV2) GetRequest() interface{} {
```
jyotimahapatra, line 120: This is the one place we had to use `interface{}` because of its use in orchestrator and http handler.

jyotimahapatra:
There are too many errors on this one to review. I'll spin up a new PR

PR #136, opened 2020-08-28
Abstract stream

jyotimahapatra:

The stream send and recv depended on the V2 references. Abstracting stream such that the upstream/client.go does not depend on V2 protos.

### DIFF
--- a/internal/app/upstream/client.go
+++ b/internal/app/upstream/client.go
@@ -195,12 +195,10 @@ func send(
 			if !ok {
 				return
 			}
-			request.UpdateNonce(sig.nonce)
-			request.UpdateVersion(sig.version)
 			// Ref: https://github.com/grpc/grpc-go/issues/1229#issuecomment-302755717
 			// Call SendMsg in a timeout because it can block in some cases.
 			err := util.DoWithTimeout(ctx, func() error {
-				return stream.SendMsg()
+				return stream.SendMsg(sig.version, sig.nonce)
 			}, callOptions.Timeout)
 			if err != nil {
 				handleError(ctx, logger, "Error in SendMsg", cancelFunc, err)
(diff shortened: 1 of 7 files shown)


internal/app/upstream/client.go
```diff
@@ -17,13 +18,13 @@ import (
...
+	ListenerTypeURL = resource.ListenerType
 	// ClusterTypeURL is the resource url for cluster
-	ClusterTypeURL = "type.googleapis.com/envoy.api.v2.Cluster"
+	ClusterTypeURL = resource.ClusterType
 	// EndpointTypeURL is the resource url for endpoints
-	EndpointTypeURL = "type.googleapis.com/envoy.api.v2.ClusterLoadAssignment"
+	EndpointTypeURL = resource.EndpointType
 	// RouteTypeURL is the resource url for route
-	RouteTypeURL = "type.googleapis.com/envoy.api.v2.RouteConfiguration"
+	RouteTypeURL = resource.RouteType
```
jessicayuen, line 27: Probably don't need these as constants anymore, would just use inline. Wdyt?

jyotimahapatra: done

internal/app/upstream/client.go
```diff
@@ -180,7 +175,7 @@ func send(
 	logger log.Logger,
 	cancelFunc context.CancelFunc,
 	request transport.Request,
```
jessicayuen, line 177: Do you think we can omit `request` from send since it's been abstracted? I know the logger uses it below, but I wonder if we can also move that to the abstracted layer (stream.go) instead.

jessicayuen: Chatted offline, request version being logged is still incremented.

jyotimahapatra: since its a private function and we use it for logging, i would leave it there. The node id is not important to log here, so i will remove it in a next PR. the other 2 values make sense to log.

Issue #125, opened 2020-08-05
Support node metadata in aggregation rules


jessicayuen:
Support the freeform 'metadata' field in the aggregation rules.

API: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/base.proto

PR #137, opened 2020-08-28
Encode Node information in separate types in aggregation rules

eapolinario:
This PR redefines how the node information is used in aggregation rules. 

So far we have relied on the presence of the `NodeFieldType` to decide which field from the envoy Node object we used for both (1) [matching](https://github.com/envoyproxy/xds-relay/blob/2d4b2e78e157222a9a8a36b4b07217a1cc10ddce/api/protos/aggregation/v1/aggregation.proto#L35-L41), and (2) [fragment generation](https://github.com/envoyproxy/xds-relay/blob/2d4b2e78e157222a9a8a36b4b07217a1cc10ddce/api/protos/aggregation/v1/aggregation.proto#L61). In this PR we define separate types for each one of the two cases (i.e. (1) and (2)).

As we spoke offline, we opted for **not** using the envoy protos, instead, we're going to maintain a mapping between the envoy objects and the objects needed to perform (1) and (2) so as to isolate ourselves from other migration issues. Concretely, the envoy v2/v3 Locality object has its properties duplicated in our copy of the `LocalityMatch` and `LocalityResultAction`.

The idea behind this refactoring is to allow for better extensibility of both the matching and fragment generation phases. In summary, the process of adding a new type of aggregation rule object consists of a 2 step process:
1. Add a pair of types to do (1) and (2) to the [aggregation rules proto](https://github.com/envoyproxy/xds-relay/blob/master/api/protos/aggregation/v1/aggregation.proto).
2. Handle the newly added types in the process the types added in the above step in the mapper (including unit tests).

A followup PR is going to follow the model proposed in this PR to add support for [node metadata](https://github.com/envoyproxy/xds-relay/issues/125).


samrabelachew (dismissed):
lgtm

example/config-files/aggregation-rules.yaml
```diff
@@ -9,8 +9,7 @@ fragments:
               - "type.googleapis.com/envoy.api.v2.RouteConfiguration"
         result:
           request_node_fragment:
-            field: 1
-            action:
+            cluster_action:
```
jyotimahapatra, line 12: i like how its a named action, instead of `field: 1`

internal/app/mapper/mapper.go
```diff
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strings"
 
+	corev2 "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
```
jyotimahapatra, line 8: we wouldn't want to import v2 proto here. The envoy v2/v3 proto usage is localized in transport package.

jyotimahapatra (commented):
The structure makes the yaml much readable. I've some concerns about using the `corev2` protos outside of transport package. Can we isolate the usage?

internal/app/mapper/mapper.go
eapolinario: I ended up defining our own `Locality` object and put it in `transport/request.go`. Each version is then responsible for doing the mapping between each version and our definition of locality. We didn't have to do that in the other uses of the mapper because all fields are defined in terms of primitive types.

Let me know if maintaining our own locality object is a reasonable compromise.

internal/app/mapper/mapper.go
```diff
@@ -124,20 +124,25 @@ func isNodeMatch(matchPredicate *matchPredicate, req transport.Request) (bool, e
...
-	case aggregationv1.NodeFieldType_NODE_LOCALITY_REGION:
-		return compare(predicate, req.GetRegion())
-	case aggregationv1.NodeFieldType_NODE_LOCALITY_ZONE:
-		return compare(predicate, req.GetZone())
-	case aggregationv1.NodeFieldType_NODE_LOCALITY_SUBZONE:
-		return compare(predicate, req.GetSubZone())
-	default:
-		return false, fmt.Errorf("RequestNodeMatch does not have a valid NodeFieldType")
+
+	// By construction only one of these is set at any point in time, so checking one by one
```
eapolinario, line 128: @jyotimahapatra , do you know if there's a way to preserve the switch-like functionality here? From a correctness perspective my comment is true, albeit less performant than the old code.

internal/app/mapper/mapper.go
```diff
@@ -402,16 +392,50 @@ func getResultFragmentFromAction(
...
+		}
+		matches = append(matches, fragment)
+	}
+
+	if len(matches) == 0 {
+		return "", fmt.Errorf("RequestNodeFragment match resulted in an empty fragment")
+	}
+
+	// N.B.: join matches using "|" to indicate they all came from the locality object.
+	return strings.Join(matches, "|"), nil
```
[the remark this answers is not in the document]
jyotimahapatra: wondering if a mapper_test should cover this?

jyotimahapatra (commented):
lgtm and few questions

internal/app/mapper/mapper.go
eapolinario: relying on short-circuiting is not better in terms of perf than what I'm proposing here.

internal/app/mapper/mapper.go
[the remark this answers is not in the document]
eapolinario: the tests I added in `mapper_test` all rely on behavior, e.g. https://github.com/envoyproxy/xds-relay/blob/31554396b7085ba18b90ffd409abd29aad7b5724/internal/app/mapper/mapper_test.go#L319.

internal/app/mapper/mapper.go
jyotimahapatra: i think what we have here works.

internal/app/mapper/mapper.go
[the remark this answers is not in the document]
jyotimahapatra: thanks for showing this

jyotimahapatra (approved):

internal/app/mapper/mapper_test.go
```diff
@@ -536,159 +527,111 @@ var negativeTests = []TableEntry{
...
 	{
 		Description: "RequestNodeMatch with exact match request node zone mismatch",
 		Parameters: []interface{}{
-			getRequestNodeExactMatch(nodeZoneField, nodezone),
-			getResultStringFragment(),
-			getDiscoveryRequestWithNode(getNode(nodeid, nodecluster, noderegion, "mismatch", nodesubzone)),
-		},
-	},
-	{
-		Description: "RequestNodeMatch with regex match request node zone mismatch",
```
jyotimahapatra, line 673: I think we removed this tests by mistake? Lets add them back

jyotimahapatra:
```	
RequestNodeMatch with node region regex does not match
RequestNodeMatch with node zone regex does not match
RequestNodeMatch with node subzone regex does not match
RequestNodeMatch with regex match request node region mismatch
RequestNodeMatch with regex match request node zone mismatch
empty node region in request predicate
empty node zone in request predicate
empty node subzone in request predicate
```

PR #138, opened 2020-08-28
Add v3 types

jyotimahapatra:
This PR adds the v3 types into transport. Since the other parts of orchestrator were abstracted, most changes are in the transport package and also adds unit tests. The next PR will add integration tests using v3 flows.


jyotimahapatra:
@jessicayuen @eapolinario This is ready to review now

internal/app/upstream/client_test.go
```diff
@@ -27,6 +30,16 @@ func TestOpenStreamShouldReturnErrorForInvalidTypeUrl(t *testing.T) {
 	assert.Nil(t, respCh)
 }
 
+func TestOpenStreamShouldReturnErrorForInvalidTypeUrlV3(t *testing.T) {
```
jyotimahapatra, line 33: i had to copy these tests because the setup is not going well with BDD style testing. I could have split the tests into v2 and v3 files, but preferred this way for readability. Let me know if you feel we should split the tests into v2 and v3 files.

internal/app/admin/http/handler_test.go
```diff
@@ -337,6 +341,145 @@ func TestAdminServer_CacheDumpHandler_EntireCache(t *testing.T) {
...
+  },
+  "Requests": [
+    {
+      "version_info": "1",
+      "type_url": "type.googleapis.com/envoy.config.listener.v3.Listener"
+    }
+  ],
+  "ExpirationTime": "`)
+		assert.Contains(t, rr.Body.String(), `cdsv3: {
+  "Resp": {
```
jessicayuen, line 449: Can we add these a fixtures in testdata? For the expirationTime, you can probably use Go templates.

internal/app/orchestrator/downstream.go
```diff
@@ -110,12 +110,7 @@ func (d *downstreamResponseMap) getAggregatedKeys(m *mapper.Mapper) (map[string]
 func (responseChannel *responseChannel) addResponse(resp transport.Response) error {
 	responseChannel.wg.Add(1)
 	defer responseChannel.wg.Done()
-	ok, err := responseChannel.watch.Send(resp)
-	if err != nil {
-		return err
-	}
```
jessicayuen, line 116: What's the reason for removing the `err` return here?

internal/app/transport/request_test.go
```diff
@@ -0,0 +1,144 @@
...
+	nodeID          = "1"
+	resourceName    = "route"
+	resourceVersion = "version"
+	cluster         = "cluster"
+	region          = "region"
+	zone            = "zone"
+	subzone         = "subzone"
+)
+
+var requestV2 = discoveryv2.DiscoveryRequest{
```
jessicayuen, line 25: May be better to structure the v2 tests in `requestv2_test.go` and the v3 ones in `requestv3_test.go` since you made the refactor here to separate the implementation from the interface.

internal/app/transport/response_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+const (
+	nonce   = "nonce"
+	typeURL = "typeUrl"
+)
+
+var responseV2 = discoveryv2.DiscoveryResponse{
```
jessicayuen, line 17: Similar to the above comment about where the tests reside.

internal/app/upstream/client_test.go
jessicayuen: I think how you structured this is OK since the client impl is shared between v2 and v3.

internal/app/orchestrator/downstream.go
jyotimahapatra: The Send function earlier used casting and returned an error when cast failed
```
func (w *watchV2) Send(s Response) (bool, error) {
	resp, ok := s.(*ResponseV2)
	if !ok {
		return false, fmt.Errorf("payload %s could not be casted to DiscoveryResponse", s)
	}
```

With the addition of VersionedRequest/Response, we got rid of casting, and there's no more error being returned by Send.
```
func (w *watchV2) Send(s Response) bool {
	select {
	case w.out <- gcpv2.PassthroughResponse{DiscoveryResponse: s.Get().V2, Request: *s.GetRequest().V2}:
		return true
	default:
		return false
	}
}
```

internal/app/admin/http/handler_test.go
jyotimahapatra: separated the fixtures

internal/app/transport/request_test.go
jyotimahapatra: I think i prefer this structure. Imagine this is the test for the interface and it ties the behavior of all impl together in one place. Let me know if you still recommend to make the change.

internal/app/admin/http/testdata/entire_cachev3_lds.json
```diff
@@ -0,0 +1,30 @@
...
+        "ControlPlane": null
+      },
+      "Requests": [
+        {
+          "version_info": "1",
+          "type_url": "type.googleapis.com/envoy.config.listener.v3.Listener"
+        }
+      ],
+      "ExpirationTime": ""
+    }
```
jessicayuen, line 30: nit: newline

internal/app/admin/http/testdata/lds_response.json
```diff
@@ -0,0 +1,34 @@
...
+      "Requests": [
+        {
+          "version_info": "1",
+          "type_url": "type.googleapis.com/envoy.api.v2.Listener"
+        }
+      ],
+      "ExpirationTime": ""
+    }
+  ]
+}
```
jessicayuen, line 34: nit: newline

jessicayuen (approved):

internal/app/admin/http/handler_test.go
```diff
@@ -138,34 +146,10 @@ func TestAdminServer_CacheDumpHandler(t *testing.T) {
...
-  "Requests": [
-    {
-      "version_info": "1",
-      "type_url": "type.googleapis.com/envoy.api.v2.Listener"
-    }
-  ],
-  "ExpirationTime": "`)
+
+	body := dateRegex.ReplaceAllString(rr.Body.String(), "\"\"")
+	filecontents, _ := ioutil.ReadFile("testdata/lds_response.json")
```
jessicayuen, line 151: might want to do an `assert.NoError(...)` here and on the below file reads.

internal/app/admin/http/testdata/lds_response.json
jyotimahapatra: i tried doing it, but the response doesnt match anymore because we do a strict check between filecontents and the response.

jessicayuen (approved):

Issue #88, opened 2020-05-18
Regex support in cache_dump admin API


LisaLudique:
See #75 for more details.

The cache_dump admin endpoint should support regex keys, i.e. dumping resources with keys that match the given regex from the cache.

PR #139, opened 2020-09-08
Cache globbing support

samrabelachew:

This change provides support for glob pattern keys in the cache endpoint as defined in https://github.com/envoyproxy/xds-relay/issues/88


samrabelachew:
@jessicayuen @eapolinario

internal/app/admin/http/handler.go
```diff
@@ -102,45 +102,73 @@ func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc
...
-		resource, err := cache.FetchReadOnly(cacheKey)
-		if err != nil {
-			fmt.Fprintf(w, "no resource for key %s found in cache.\n", cacheKey)
-			return
-		}
-		resourceString, err := resourceToString(resource)
-		if err != nil {
-			w.WriteHeader(http.StatusInternalServerError)
-			fmt.Fprintf(w, "unable to convert resource to string.\n")
+		// If wildcard suffix provided, output all cache entries that match given prefix.
```
eapolinario, line 139: What do you think of the idea of combining these two if-blocks? We can detect the case of `cacheKey` being empty string and replace with the value `"*"`, then we could handle both cases with a single if-block.

eapolinario (commented):
How about we are even more explicit about the type of matching we're allowing with this PR? More concretely, how about we call out that this is really adding prefix matching to the cache dump endpoint (and the convention being to add a single asterisk at the end of the url param)? This way we explicitly disallow other globbing usecases, like multiple uses of the asterisk in different parts of the cache key (e.g., what if someone wanted to list all keys of the format `abc*e*`?)

Also, maybe not in this PR, but we should document this endpoint as it's very useful to inspect what's happening inside xds-relay. Maybe add a section in the main readme with an example?

internal/app/admin/http/handler.go
samrabelachew: Sure, I still want to add an extra if to avoid checking for prefixes in the print total cache case to avoid unnecessary steps.

samrabelachew:
> How about we are even more explicit about the type of matching we're allowing with this PR? More concretely, how about we call out that this is really adding prefix matching to the cache dump endpoint (and the convention being to add a single asterisk at the end of the url param)? This way we explicitly disallow other globbing usecases, like multiple uses of the asterisk in different parts of the cache key (e.g., what if someone wanted to list all keys of the format `abc*e*`?)
> 
> Also, maybe not in this PR, but we should document this endpoint as it's very useful to inspect what's happening inside xds-relay. Maybe add a section in the main readme with an example?

There is documentation of it in the xDS Relay [spec](https://docs.google.com/document/d/1X9fFzqBZzrSmx2d0NkmjDW2tc8ysbLQS9LaRQRxdJZU/edit#heading=h.wg5z5aoqsa70), but I'll be sure to add a section in a separate PR

internal/app/admin/http/handler.go
```diff
@@ -102,45 +102,62 @@ func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc
...
-				if err == nil {
-					resourceString, err := resourceToString(resource)
-					if err == nil {
-						fmt.Fprintf(w, "%s: %s\n", key, resourceString)
+
+			// Find keys that match prefix of wildcard
+			if containsWildcardSuffix {
+				for potentialMatchKey := range keys {
+					if !strings.HasPrefix(potentialMatchKey, rootCacheKeyName) {
+						delete(keys, potentialMatchKey)
```
jyotimahapatra, line 143: is it safe to delete from a list/map while iterating on it?

internal/app/admin/http/handler.go
```diff
@@ -102,45 +102,62 @@ func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc
 	}
 }
 
-// TODO(lisalu): Support dump of matching resources when cache key regex is provided.
+func printCacheEntry(key string, cache cache.ReadOnlyCache, w http.ResponseWriter) {
```
jyotimahapatra, line 105: i would suggest changing the signature to `key string[]` and condense the multiple return paths in `cacheDumpHandler`

internal/app/admin/http/handler.go
samrabelachew: I also was concerned about this, but I think it is safe. Golang's [docs](https://golang.org/ref/spec#For_statements) imply that it is okay to delete entries while iterating through a map in rule three of the "For statements with range clause" section. Let me know if I misinterpreted it though!

eapolinario: it's a very unusual design decision, but yeah, it totally works.

eapolinario (approved):
LGTM.

internal/app/admin/http/handler_test.go
```diff
@@ -361,6 +387,301 @@ func TestAdminServer_CacheDumpHandler_EntireCacheV3(t *testing.T) {
...
+
+		handler.ServeHTTP(rr, req)
+		assert.Equal(t, http.StatusOK, rr.Code)
+
+		body := dateRegex.ReplaceAllString(rr.Body.String(), "\"\"")
+		filecontentsCds, err := ioutil.ReadFile("testdata/entire_cachev2_cds.json")
+		assert.NoError(t, err)
+		filecontentsLds, err := ioutil.ReadFile("testdata/entire_cachev2_lds.json")
+		assert.NoError(t, err)
+		assert.Contains(t, body, string(filecontentsCds))
```
eapolinario, line 482: this will only match the a single entry, right? Can you add a test case where we match multiple keys and we assert the contents of body exactly?

eapolinario (changes requested):

internal/app/admin/http/handler_test.go
samrabelachew: Sorry, can you clarify what you mean here? This matches the generated keys `test_cds` and `test_lds` because they both begin with "t", "tes", and "test".

internal/app/admin/http/handler.go
```diff
@@ -109,52 +109,73 @@ func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc
...
-
-		resp.Cache = append(resp.Cache, resourceToPayload(cacheKey, resource)...)
-		resourceString, err := stringify.InterfaceToString(resp)
-		if err != nil {
-			w.WriteHeader(http.StatusInternalServerError)
-			fmt.Fprintf(w, "unable to convert resource to string.\n")
-			return
+			// Add matched keys to print slice
+			for key := range keys {
+				keysToPrint = append(keysToPrint, key)
```
jyotimahapatra, line 172: instead of deleting keys earlier, we can instead add keys to `keysToPrint` directly.

internal/app/admin/http/handler_test.go
eapolinario: indeed, thanks for clarifying this. Disregard my comment then. This looks good.

internal/app/admin/http/handler.go
```diff
@@ -109,52 +109,72 @@ func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc
...
+					}
+				}
+				if len(keysToPrint) == 0 {
+					w.WriteHeader(http.StatusNotFound)
+					fmt.Fprintf(w, "no resource for key %s found in cache.\n", cacheKey)
+					return
+				}
+			} else {
+				// Add matched keys to print slice
+				for key := range allKeys {
```
eapolinario, line 169: do you think it's worth special-casing the case of `allKeys` being empty?

samrabelachew: If you mean creating a printed response for the empty case, I can add an output statement like "cache is empty, zero entries found". If I do that though, I feel like I should also incorporate that logic for when the URL is `/cache/*`.

samrabelachew: Offline, we discussed this more and decided to move forward with no output in the empty return cases as it avoids adding extra messages that may need to be filtered out. Instead, I'll explore looking into adding cases like these as debug logs in the future.

eapolinario (approved):

internal/app/admin/http/handler.go
```diff
@@ -109,52 +109,67 @@ func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc
...
 		c := orchestrator.Orchestrator.GetReadOnlyCache(*o)
-		resp := marshallableCache{}
+		var keysToPrint []string
+
+		// If wildcard suffix provided, output all cache entries that match given prefix.
 		// If no key is provided, output the entire cache.
-		if cacheKey == "" {
-			keys, err := orchestrator.Orchestrator.GetDownstreamAggregatedKeys(*o)
+		containsWildcardSuffix := strings.HasSuffix(cacheKey, "*")
+		if cacheKey == "" || containsWildcardSuffix {
```
jessicayuen, line 145: What do you think of making a separate function:
```
// hasWildcardSuffix returns whether the supplied key contains an empty string or a * suffix.
// Return true in these scenarios and the prefix.
func hasWildcardSuffix(key string) (bool, string) {
    return if key == "" || strings.HasSuffix(key, "*"), strings.TrimSuffix(key, "*")
}
```

I think this way it can help get rid of some of the nested ifs between the two pathways and treat them equally, just testing for the suffix.

samrabelachew: I think I'm misinterpreting your function here, but I don't understand how it removes the need of two pathways. The original goal of having two pathways was to just avoid doing prefix checks for the "print entire cache" case since it was unnecessary. I don't think it's an expensive operation, so I can go ahead and remove the nested ifs and treat is as one case. Is this what you mean?

```
               if cacheKey == "" || containsWildcardSuffix {
			// Retrieve all keys
			allKeys, err := orchestrator.Orchestrator.GetDownstreamAggregatedKeys(*o)
			if err != nil {
				w.WriteHeader(http.StatusInternalServerError)
				fmt.Fprintf(w, "error in getting cache keys: %s", err.Error())
				return
			}

			// Find keys that match prefix of wildcard
			rootCacheKeyName := strings.TrimSuffix(cacheKey, "*")
			for potentialMatchKey := range allKeys {
				if strings.HasPrefix(potentialMatchKey, rootCacheKeyName) {
					keysToPrint = append(keysToPrint, potentialMatchKey)
				}
			}
		} ....
```

internal/app/admin/http/handler.go
```diff
@@ -109,52 +109,67 @@ func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc
...
 
-// TODO(lisalu): Support dump of matching resources when cache key regex is provided.
+func printCacheEntries(keys []string, cache cache.ReadOnlyCache, w http.ResponseWriter) {
+	resp := marshallableCache{}
+	for _, key := range keys {
+		resource, err := cache.FetchReadOnly(key)
+		if err != nil {
+			w.WriteHeader(http.StatusNotFound)
+			fmt.Fprintf(w, "no resource for key %s found in cache.\n", key)
+			continue
```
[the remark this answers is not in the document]
jessicayuen: For public prosperity, we chatted offline and decided to skip the key but not return 404 or log an error. Despite multiple reads this should be OK since it's all an atomic action for this admin API.

internal/app/admin/http/handler.go
jyotimahapatra: My version of a shorter function 
```
func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
		cacheKey := getParam(req.URL.Path)
		c := orchestrator.Orchestrator.GetReadOnlyCache(*o)
		var keysToPrint []string

		// If wildcard suffix provided, output all cache entries that match given prefix.
		// If no key is provided, output the entire cache.
               if cacheKey == "" {
                 cacheKey = "*"
               }
		containsWildcardSuffix := strings.HasSuffix(cacheKey, "*")
		if containsWildcardSuffix {
			// Retrieve all keys
			allKeys, err := orchestrator.Orchestrator.GetDownstreamAggregatedKeys(*o)
			if err != nil {
				w.WriteHeader(http.StatusInternalServerError)
				fmt.Fprintf(w, "error in getting cache keys: %s", err.Error())
				return
			}

                       rootCacheKeyName := strings.TrimSuffix(cacheKey, "*")
		       for potentialMatchKey := range allKeys {
			if strings.HasPrefix(potentialMatchKey, rootCacheKeyName) {
					keysToPrint = append(keysToPrint, potentialMatchKey)
				}
			}
		} else {
			// Otherwise return the cache entry corresponding to the given key.
			keysToPrint = []string{cacheKey}
		}
		printCacheEntries(keysToPrint, c, w)
	}
}
```

jessicayuen: @samrabelachew I just meant to get rid of [this](https://github.com/envoyproxy/xds-relay/pull/139/files#diff-29e3fee34912e53bdb13be1567ad0307R155-L133) nested block to improve readability, rather than perf gains. There's multiple ways to go about it including the one Jyoti listed, I'm ok with either.

internal/app/admin/http/handler.go
```diff
@@ -109,52 +109,62 @@ func configDumpHandler(bootstrapConfig *bootstrapv1.Bootstrap) http.HandlerFunc
 	}
 }
 
-// TODO(lisalu): Support dump of matching resources when cache key regex is provided.
+func printCacheEntries(keys []string, cache cache.ReadOnlyCache, w http.ResponseWriter) {
+	resp := marshallableCache{}
+	for _, key := range keys {
+		resource, err := cache.FetchReadOnly(key)
+		if err != nil {
```
jessicayuen, line 116: should be `err == nil` I think

jessicayuen: or better,
```
if err != nil {
    resp.Cache = append(resp.Cache, resourceToPayload(key, resource)...)
}
```

jessicayuen (commented):
LGTM pending the above comment. Thanks!

internal/app/admin/http/handler.go
samrabelachew: @jessicayuen  err equaling nil means we have no error right (i.e. the fetch returned successfully)? If it returned successfully, we should append the resource.

samrabelachew: i think you mean this?
```
if err == nil {
    resp.Cache = append(resp.Cache, resourceToPayload(key, resource)...)
}
```

jessicayuen: yep, derp my mistake. How you had it is fine too.

jessicayuen (approved):

PR #142, opened 2020-09-11
Improve integration test

jyotimahapatra:
The integration tests were not stopping envoy between each version. 
The test checks that the hello world http server is able to respond. Since envoy was not restarted between versions, the tests were still using v2 configurations and passing.

Also we've identified the reason of false positive. We were not checking the output after starting envoy, in case envoy never starts, the buffer is empty.


integration/e2e_test.go
```diff
@@ -90,9 +99,14 @@ func TestSnapshotCacheSingleEnvoyAndXdsRelayServer(t *testing.T) {
 				}, 1*time.Second, 100*time.Millisecond).Should(gomega.Equal(nListeners))
 			}
 
+			// If envoy never starts, the buffer is empty
+			assert.NotEmpty(t, envoyLogsBuffer.String())
```
jyotimahapatra, line 103: This condition could have prevented the false positive.

integration/e2e_test.go
```diff
@@ -90,9 +99,14 @@ func TestSnapshotCacheSingleEnvoyAndXdsRelayServer(t *testing.T) {
...
 			// TODO(https://github.com/envoyproxy/xds-relay/issues/66): figure out a way to only only copy
 			// envoy logs in case of failures.
 			testLogger.With("envoy_logs", envoyLogsBuffer.String()).Debug(ctx, "captured envoy logs")
+
+			err := stopEnvoy(ctx, pid)
```
jessicayuen, line 108: Should this be a separate test? I think we want to test the codepath where envoy is receiving continuous updates without restarting.

jyotimahapatra: We have to start a new envoy instance to allow a new bootstrap. We can use separate port numbers to start a new instance of envoy, or shutdown the old one and start a new one. I like shutting and restarting more since its clean and will surface more bugs.

internal/app/cache/cache.go
```diff
@@ -233,7 +233,7 @@ func (c *cache) DeleteRequest(key string, req transport.Request) error {
 	c.cache.Add(key, resource)
 	c.logger.With(
 		"aggregated_key", key,
-		"node_id", req.GetNodeID,
+		"node_id", req.GetNodeID(),
```
eapolinario, line 236: wow. What does this even print if it's not a function call? The address of the function?

integration/e2e_test.go
eapolinario: I don't understand how this helps. The envoy logs are only for debugging in case the tests fails.

integration/e2e_test.go
eapolinario: @jessicayuen , if we want to test the scenario where the control plane sends v2 and v3 interspersed we should create a different test that tests this specifically.

@jyotimahapatra , now that we're killing envoy manually, can you try removing the Linux-only check I added and check if the tests fail? You can inspect if there's any dangling process after tests run by inspecting the docker container (run `docker run --rm -it --pid host jess/htop` in a separate window and search for the envoy process)

integration/e2e_test.go
jyotimahapatra: yes...additionally when envoy is not able to start up the logs are empty.

internal/app/cache/cache.go
jyotimahapatra: it printed the function name  😄

integration/e2e_test.go
jyotimahapatra: I verified that the verification steps mentioned work. Also, i think we have safety due to these:
1. If kill gets stuck for any reason, the test will timeout .
2. If kill gives back an error, the test fails due to err check.

eapolinario (approved):
LGTM.

PR #146, opened 2020-09-16
Evict cache on upstream failure

jyotimahapatra:
When upstream control plane rotates, the grpc client has to make a new connection. Today we recognize an issue in the stream and dont retry. We should at least close the downstream sidecar streams, so that we get some semblance of retry from envoy sidecars. This could cause thundering herd, but this is a stopgap until the next PR fixes the retry.

### DIFF
--- a/internal/app/orchestrator/orchestrator.go
+++ b/internal/app/orchestrator/orchestrator.go
@@ -280,11 +280,11 @@ func (o *orchestrator) watchUpstream(
 		case x, more := <-responseChannel:
 			if !more {
 				// A problem occurred fetching the response upstream, log retry.
-				// TODO implement retry/back-off logic on error scenario.
-				// https://github.com/envoyproxy/xds-relay/issues/68
 				o.logger.With("aggregated_key", aggregatedKey).Error(ctx, "upstream error")
 				metrics.OrchestratorWatchErrorsSubscope(o.scope, aggregatedKey).Counter(metrics.ErrorUpstreamFailure).Inc(1)
 
+				// TODO implement retry/back-off logic on error scenario.
+				// https://github.com/envoyproxy/xds-relay/issues/68
 				f, err := o.cache.Fetch(aggregatedKey)
 				if err == nil {
 					o.onCacheEvicted(aggregatedKey, *f)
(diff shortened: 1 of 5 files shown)


internal/app/orchestrator/orchestrator.go
```diff
@@ -280,10 +280,16 @@ func (o *orchestrator) watchUpstream(
 		case x, more := <-responseChannel:
 			if !more {
 				// A problem occurred fetching the response upstream, log retry.
-				// TODO implement retry/back-off logic on error scenario.
-				// https://github.com/envoyproxy/xds-relay/issues/68
 				o.logger.With("aggregated_key", aggregatedKey).Error(ctx, "upstream error")
 				metrics.OrchestratorWatchErrorsSubscope(o.scope, aggregatedKey).Counter(metrics.ErrorUpstreamFailure).Inc(1)
+
+				// TODO implement retry/back-off logic on error scenario.
+				// https://github.com/envoyproxy/xds-relay/issues/68
```
jessicayuen, line 287: Should we add a comment that also says something like
`Currently we force close all downstream connections, forcing downstream clients to issue a retry.` or we could just add it to the issue too.

internal/app/transport/streamv3.go
```diff
@@ -10,20 +13,28 @@ var _ Stream = &streamv3{}
...
 }
 
 func (s *streamv3) SendMsg(version string, nonce string) error {
 	msg := s.initialRequest.GetRaw().V3
 	msg.VersionInfo = version
 	msg.ResponseNonce = nonce
+	s.logger.With(
+		"request_type", msg.GetTypeUrl(),
+		"request_version", msg.GetVersionInfo(),
+		"request_msg", msg.ResourceNames,
```
jessicayuen, line 35: nit: should this be `request_resource_names` since the message is more than the resource names?

jessicayuen (dismissed):
LGTM, couple non-blocking comments

internal/app/orchestrator/orchestrator.go
jyotimahapatra: added it to the issue.

internal/app/transport/streamv3.go
jyotimahapatra: makes sense. I dont know why i named it request_msg previously.

jyotimahapatra:
@jessicayuen I took care of the comments. Can you approve again.

jessicayuen (approved):

PR #148, opened 2020-09-17
Add stopgap for unaggregated keys

jessicayuen:
Log error and close the watch.

For longer term, there is an open issue to support unaggregated keys.

### DIFF
--- a/internal/app/orchestrator/orchestrator.go
+++ b/internal/app/orchestrator/orchestrator.go
@@ -8,7 +8,6 @@ package orchestrator
 
 import (
 	"context"
-	"fmt"
 	"sync"
 
 	bootstrapv1 "github.com/envoyproxy/xds-relay/pkg/api/bootstrap/v1"
@@ -25,10 +24,6 @@ import (
 
 const (
 	component = "orchestrator"
-
-	// unaggregatedPrefix is the prefix used to label discovery requests that
-	// could not be successfully mapped to an aggregation rule.
-	unaggregatedPrefix = "unaggregated_"
 )
 
 // Orchestrator has the following responsibilities:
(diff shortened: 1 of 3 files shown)


internal/app/orchestrator/orchestrator.go
```diff
@@ -137,22 +132,17 @@ func (o *orchestrator) CreateWatch(req transport.Request) (transport.Watch, func
...
 			"error", err,
 			"request_type", req.GetTypeURL(),
-			"request", req,
-		).Warn(ctx, "failed to map to aggregated key")
-		// Mimic the aggregated key.
-		// TODO (https://github.com/envoyproxy/xds-relay/issues/56). Can we
-		// condense this key but still make it granular enough to uniquely
-		// identify a request?
-		if req.GetRaw().V2 != nil {
-			aggregatedKey = fmt.Sprintf("%s%s", unaggregatedPrefix, req.GetRaw().V2)
```
jyotimahapatra, line 152: Can you check if the `GetRaw` function is still used in other places and remove it if its not.

internal/app/orchestrator/orchestrator_test.go
```diff
@@ -177,6 +177,35 @@ func TestGoldenPath(t *testing.T) {
...
+		},
+	)
+	// assert this request will not map to an aggregated key.
+	_, err := mapper.GetKey(req)
+	assert.Error(t, err)
+
+	respChannel, _ := orchestrator.CreateWatch(req)
+	testutils.AssertCounterValue(t, mockScope.Snapshot().Counters(),
+		fmt.Sprintf("mock_orchestrator.watch.errors.unaggregated_key"), 1)
+	assert.NotNil(t, respChannel)
```
jyotimahapatra, line 205: is it possible to check the watch is closed?

jyotimahapatra (commented):
lgtm, some questions.

internal/app/orchestrator/orchestrator.go
jessicayuen: Still used here unfortunately: https://github.com/envoyproxy/xds-relay/blob/85bc48c1fffae9a13e7c41a5d572d6eeec48496a/internal/app/orchestrator/orchestrator.go#L163 :(

jyotimahapatra: i think this is a mistake from my previous refactoring. There should be handling for both v2/v3 here. I will fix it in a separate PR

jyotimahapatra (approved):

PR #150, opened 2020-09-18
Add upstream connectivity stat

jyotimahapatra:
We need a way to know quickly the connectivity status between xds relay and the upstream control plane. This PR adds a gauge to reflect the last known connectivity state.

Lyft internal [dashboard](https://lyft.wavefront.com/u/vRgKSC6G8V?t=lyft).

### DIFF
--- a/internal/app/upstream/client.go
+++ b/internal/app/upstream/client.go
@@ -19,7 +19,6 @@ import (
 	"github.com/envoyproxy/xds-relay/internal/pkg/util"
 	"github.com/uber-go/tally"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/connectivity"
 )
 
 // UnsupportedResourceError is a custom error for unsupported typeURL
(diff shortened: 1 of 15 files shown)


samrabelachew:
lgtm

internal/app/upstream/client.go
```diff
@@ -289,3 +291,19 @@ func shutDown(ctx context.Context, conn *grpc.ClientConn) {
 func (e *UnsupportedResourceError) Error() string {
 	return fmt.Sprintf("Unsupported resource typeUrl: %s", e.TypeURL)
 }
+
+func updateConnectivityMetric(ctx context.Context, conn *grpc.ClientConn, scope tally.Scope) {
+	connectedStat := "connected"
+	for {
```
jessicayuen, line 297: Should we put a sleep here to poll every 1s instead? or is this blocking on `WaitForStateChange`?

jyotimahapatra: its a blocked channel, we woudn;t need to poll.

internal/app/upstream/client.go
```diff
@@ -289,3 +291,19 @@ func shutDown(ctx context.Context, conn *grpc.ClientConn) {
...
+		isChanged := conn.WaitForStateChange(ctx, conn.GetState())
+		// Based on the grpc implementation, isChanged is false only when ctx expires
+		// https://github.com/grpc/grpc-go/blob/0f7e218c2cf49c7b0ca8247711b0daed2a07e79a/clientconn.go#L509-L510
+		// We set an unusually high value in case of ctx expires to show the connection is dead.
+		if !isChanged {
+			scope.Gauge(connectedStat).Update(100)
```
jessicayuen, line 303: Should this be 0 to follow envoy's 0 and 1 connected states?

jyotimahapatra: I thought about it. In the previous commits in this PR i made 0 as a representative of ctx expire and made mappings for each other enum. This complexity was due to the fact that the `connectivity.State` enum started at 0. I preferred to tradeoff the extra code with the assumption to map grpc connectivity state with the gauge value to keep the code readable and use an abnormally high number for ctx cancellation state.

internal/app/upstream/client.go
```diff
@@ -289,3 +291,19 @@ func shutDown(ctx context.Context, conn *grpc.ClientConn) {
 func (e *UnsupportedResourceError) Error() string {
 	return fmt.Sprintf("Unsupported resource typeUrl: %s", e.TypeURL)
 }
+
+func updateConnectivityMetric(ctx context.Context, conn *grpc.ClientConn, scope tally.Scope) {
+	connectedStat := "connected"
```
jessicayuen, line 296: can you unify this in https://github.com/envoyproxy/xds-relay/blob/master/internal/app/metrics/metrics.go?

jessicayuen (approved):

Issue #118, opened 2020-07-29
Shutdown blocks waiting for all clients to be disconnected


eapolinario:
Currently, during graceful shutdown, we wait for all clients to be disconnected before killing the server.

If we send one of the [two signals](https://github.com/envoyproxy/xds-relay/blob/master/internal/app/server/server.go#L154) to the server process the graceful shutdown process blocks on the call to [`server.GracefulStop`](https://github.com/envoyproxy/xds-relay/blob/master/internal/app/server/server.go#L164). Here's what we see in the logs:
```
{"level":"info","ts":1595981344.7454,"msg":"received interrupt signal:%!(EXTRA string=interrupt)","json":{}}
{"level":"info","ts":1595981344.745436,"msg":"initiating admin server shutdown","json":{}}
{"level":"info","ts":1595981344.745485,"msg":"initiating grpc graceful stop","json":{}}
```

PR #153, opened 2020-09-23
Shutdown bug

samrabelachew:
Resolving Issue https://github.com/envoyproxy/xds-relay/issues/118

To keep server from hanging after sending SIGINT/SIGTERM, cancel global context before calling gracefulStop()


jessicayuen (approved):
This is awesome. Thanks for the fix.

PR #162, opened 2020-10-10
Add pprof endpoints

jyotimahapatra:
Based on https://golang.org/src/net/http/pprof/pprof.go we can add handlers to serve profiles of the currently executing program.
It will help us debug perf issues.


eapolinario (commented):
Can you add a section in the README explaining these new endpoints?

jyotimahapatra:
I will do that after experimenting how these work. Right now I'm just adding the endpoints which gives us the ability to experiment with them and add docs later.

eapolinario (approved):

PR #163, opened 2020-10-13
Fix flakey tests

jyotimahapatra:
In the upstream client tests, the `defer` way of closing goroutines was turning out to be flakey. By the time the go-leak ran the goroutines were not always done.
Instead we have to explicitly wait for each test cleanup before the go-leak ran. In go tests there is no way to add a per test teardown hook.

Verfied each test with `-count=100`


eapolinario (commented):
Does this mean that using `defer` in tests is not recommended? This seems a pretty serious issue.

jyotimahapatra:
defer works in a sequence and the sequence has to be right for the test setup. In order to guarantee the tearup sequence, i've made the context cancellations and channel closures explicit. `defer` will work if they are specified in the right order.

eapolinario (approved):
Being more explicit is better IMO, even at the expense of less clarity.

PR #166, opened 2020-10-19
Add more client stats/logs per aggregated key

jyotimahapatra:
While rolling out xdsrelay in staging, we noticed difficulty in debugging a cache drift between upstream control plane and xdsrelay.
It was hard to associate errors in xdsrelay with the per service cache drift. It would be easy to track errors if we had granular per key stats.

### DIFF
--- a/internal/app/upstream/client.go
+++ b/internal/app/upstream/client.go
@@ -49,7 +49,7 @@ type Client interface {
 	// The shutdown function represents the intent that a stream is supposed to be closed.
 	// All goroutines that depend on the ctx object should consider ctx.Done to be related to shutdown.
 	// All such scenarios need to exit cleanly and are not considered an erroneous situation.
-	OpenStream(transport.Request) (<-chan transport.Response, func())
+	OpenStream(transport.Request, string) (<-chan transport.Response, func())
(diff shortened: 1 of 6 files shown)


internal/app/upstream/client.go
```diff
@@ -134,11 +134,11 @@ func New(
 	}, nil
 }
 
-func (m *client) OpenStream(request transport.Request) (<-chan transport.Response, func()) {
+func (m *client) OpenStream(request transport.Request, aggregatedKey string) (<-chan transport.Response, func()) {
```
jessicayuen, line 137: can we just run the request through the mapper rather than modifying the interface?

jyotimahapatra: i think that cost is not necessary compared to the interface change

jyotimahapatra: i understand it looks awkward to introduce the param for just logging...i also think introducing dependency on mapper is similarly awkward. The client is supposed to be low level and should not contain the dependencies from orchestrator

jessicayuen (dismissed):

jessicayuen (approved):

PR #169, opened 2020-10-21
Add endpoints for get keys and eds

jyotimahapatra:
We are creating a inhouse monitor that will leverage these endpoints and compare xdsrelay and upstream cache.


internal/app/admin/http/key.go
```diff
@@ -0,0 +1,44 @@
+package handler
```
jessicayuen, line 1: nit: `internal/app/admin/http/cache.go`?

internal/app/admin/http/key.go
```diff
@@ -0,0 +1,44 @@
...
+
+func keyDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		allKeys, err := orchestrator.Orchestrator.GetDownstreamAggregatedKeys(*o)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			errMessage, _ := stringify.InterfaceToString(&marshallable.Error{
+				Message: fmt.Sprintf("error in getting cache keys: %s", err.Error()),
+			})
+			_, _ = w.Write([]byte(errMessage))
```
jessicayuen, line 20: Similar comment with err handling as the EDS one.

pkg/marshallable/keys.go
```diff
@@ -0,0 +1,6 @@
+package marshallable
```
jessicayuen, line 1: `cache.go`?

internal/app/admin/http/eds.go
```diff
@@ -0,0 +1,84 @@
...
+
+func edsDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		edsKey := getParam(req.URL.Path)
+		if edsKey == "" {
+			w.WriteHeader(http.StatusBadRequest)
+			s, _ := stringify.InterfaceToString(&marshallable.Error{
+				Message: "Empty key",
+			})
+			_, _ = w.Write([]byte(s))
```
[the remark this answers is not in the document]
jyotimahapatra: the linter complains about unhandled errors and the header already says bad request.

internal/app/admin/http/handler.go
```diff
@@ -58,6 +58,18 @@ func getHandlers(bootstrap *bootstrapv1.Bootstrap,
...
+			"print the eds payload for a particular key. usage: `/eds/<key>`",
+			edsDumpHandler(orchestrator),
+			true,
+		},
+		{
+			"/keys",
+			"print all keys",
+			keyDumpHandler(orchestrator),
+			true,
+		},
```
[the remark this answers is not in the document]
jyotimahapatra: makes sense

internal/app/admin/http/key.go
jyotimahapatra: here the header is already 500, there would be no difference in making the check

jessicayuen (approved):

PR #170, opened 2020-10-22
Added admin handler to clear cache entries

samrabelachew:

Created handler to allow user to clear a cache entry given a key. Shares same capabilities of wildcard suffix as cache dump handler


internal/app/cache/cache.go
```diff
@@ -219,14 +222,14 @@ func (c *cache) AddRequest(key string, req transport.Request) error {
 func (c *cache) DeleteRequest(key string, req transport.Request) error {
 	c.cacheMu.Lock()
 	defer c.cacheMu.Unlock()
-	metrics.CacheAddRequestSubscope(c.scope, key).Counter(metrics.CacheDeleteAttempt).Inc(1)
+	metrics.CacheDeleteRequestSubscope(c.scope, key).Counter(metrics.CacheDeleteAttempt).Inc(1)
```
samrabelachew, line 225: Changed the metrics for `DeleteRequest` because I think it was a mistake to use `CacheAddRequestSubscope `

samrabelachew:
@jyotimahapatra @jessicayuen

internal/app/admin/http/cache_test.go
```diff
@@ -699,6 +699,332 @@ func TestAdminServer_CacheDumpHandler_WildcardSuffix_NotFound(t *testing.T) {
 	}
 }
 
+func testAdminServerClearCacheHelper(t *testing.T, urls []string) {
```
chuang-lyft, line 702: Seems that there are some code duplication between these 3 functions. The main difference is the TypeUrl. Could some of the logic be consolidated into a "base" test function:

samrabelachew: I think the challenge is that it isn't just the TypeUrl. Since these tests check between v2 and v3, most variables, while having the same function, are still different types. I do think I can consolidate one of these tests a bit more, and will do so in next commit

internal/app/admin/http/cache.go
```diff
@@ -190,6 +185,32 @@ type marshallableCache struct {
...
+func getRelevantKeys(o *orchestrator.Orchestrator, inputKey string, w http.ResponseWriter) ([]string, error) {
+	var relevantKeys []string
+	// If wildcard suffix provided, retrieve all cache keys that match the given prefix.
+	// If no key is provided, retrieve all keys.
+	if hasWildcardSuffix(inputKey) {
+		// Retrieve all keys
+		allKeys, err := orchestrator.Orchestrator.GetDownstreamAggregatedKeys(*o)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, "error in getting cache keys: %s", err.Error())
```
jyotimahapatra, line 197: we should be sending a. marshallableobject back. https://github.com/envoyproxy/xds-relay/blob/master/pkg/marshallable/error.go

internal/app/admin/http/cache.go
```diff
@@ -210,6 +231,15 @@ func printCacheEntries(keys []string, cache cache.ReadOnlyCache, w http.Response
 	}
 }
 
+func clearCacheEntries(keys []string, cache cache.Cache, w http.ResponseWriter) {
+	for _, key := range keys {
+		err := cache.DeleteKey(key)
+		if err != nil {
+			fmt.Fprintln(w, err.Error())
```
jyotimahapatra, line 238: similar to above return a marshallable response back

internal/app/cache/cache.go
```diff
@@ -236,7 +239,21 @@ func (c *cache) DeleteRequest(key string, req transport.Request) error {
...
+func (c *cache) DeleteKey(key string) error {
+	c.cacheMu.Lock()
+	defer c.cacheMu.Unlock()
+	metrics.CacheDeleteKeySubscope(c.scope, key).Counter(metrics.CacheDeleteKeyAttempt).Inc(1)
+	_, found := c.cache.Get(key)
+	if !found {
+		metrics.CacheDeleteKeySubscope(c.scope, key).Counter(metrics.CacheDeleteKeyError).Inc(1)
+		return fmt.Errorf("unable to delete entry for nonexistent key: %s", key)
+	}
+	c.cache.Remove(key)
```
jyotimahapatra, line 255: i think we need to cancel the downstream watches and upstream stream when this happens

internal/app/admin/http/cache.go
```diff
@@ -108,12 +108,10 @@ func edsDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 
 func keyDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
-		allKeys, err := orchestrator.Orchestrator.GetDownstreamAggregatedKeys(*o)
-		if err != nil {
+		allKeys, marshallableError := orchestrator.Orchestrator.GetDownstreamAggregatedKeys(*o)
+		if marshallableError.Message != "" {
```
jyotimahapatra, line 112: does `error` interface have a `Message` field?

internal/app/admin/http/cache.go
```diff
@@ -151,31 +147,26 @@ func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
...
+			c := orchestrator.Orchestrator.GetCache(*o)
+			keysToClear, err := getRelevantKeys(o, cacheKey, w)
+			if err.Message == "" {
+				clearCacheEntries(keysToClear, c, o, w)
 			}
 		} else {
-			// Otherwise return the cache entry corresponding to the given key.
-			keysToPrint = []string{cacheKey}
+			w.WriteHeader(http.StatusMethodNotAllowed)
+			fmt.Fprintf(w, "Only POST is supported\n")
```
jyotimahapatra, line 168: can we return an `marshallable.Error` instead of writing raw string into the response?

internal/app/admin/http/cache.go
```diff
@@ -210,6 +228,17 @@ func printCacheEntries(keys []string, cache cache.ReadOnlyCache, w http.Response
 	}
 }
 
+func clearCacheEntries(keys []string, cache cache.Cache, o *orchestrator.Orchestrator, w http.ResponseWriter) {
+	for _, key := range keys {
+		resource, err := cache.DeleteKey(key)
```
jyotimahapatra, line 233: is there necessity for doing this operation inside a lock?

internal/app/cache/cache.go
```diff
@@ -236,10 +240,33 @@ func (c *cache) DeleteRequest(key string, req transport.Request) error {
 		"node_id", req.GetNodeID(),
 		"request_type", req.GetTypeURL(),
 	).Debug(context.Background(), "request deleted")
-	metrics.CacheAddRequestSubscope(c.scope, key).Counter(metrics.CacheDeleteSuccess).Inc(1)
+	metrics.CacheDeleteRequestSubscope(c.scope, key).Counter(metrics.CacheDeleteSuccess).Inc(1)
 	return nil
 }
 
+func (c *cache) DeleteKey(key string) (Resource, marshallable.Error) {
```
jyotimahapatra, line 247: The `marshallable.Error` should be used at the periphery of webapi calls. The rest of the code should return `error`.

internal/app/admin/http/cache.go
jyotimahapatra: would it be better if this operation was part of orchestrator?

internal/app/orchestrator/orchestrator.go
```diff
@@ -52,11 +53,15 @@ type Orchestrator interface {
 	// open channels.
 	shutdown(ctx context.Context)
 
+	GetCache() cache.Cache
+
 	GetReadOnlyCache() cache.ReadOnlyCache
 
-	GetDownstreamAggregatedKeys() (map[string]bool, error)
+	GetDownstreamAggregatedKeys() (map[string]bool, marshallable.Error)
```
jyotimahapatra, line 60: the orchestrator interface should return an error following idiomatic go patterns. The json formatting is only for the purpose of the webapi.

internal/app/orchestrator/orchestrator.go
```diff
@@ -385,6 +398,11 @@ func (o *orchestrator) onCacheEvicted(key string, resource cache.Resource) {
 	metrics.OrchestratorCacheEvictSubscope(o.scope, key).Counter(
 		metrics.OrchestratorOnCacheEvictedRequestCount).Inc(int64(len(resource.Requests)))
 	o.logger.With("aggregated_key", key).Debug(context.Background(), "cache eviction called")
+	o.DeleteAll(key, resource)
```
jyotimahapatra, line 401: was this a bug in current flow?

samrabelachew: I reverted the change. It is no longer relevant since finding out onCacheEvicted is called for all cache removals.

internal/app/cache/cache.go
samrabelachew: c.cache.Remove() also runs the onEvicted() func internally, which deletes the downstream watches and upstream stream since it is currently set to onCacheEvicted():
https://github.com/envoyproxy/xds-relay/blob/master/internal/app/orchestrator/orchestrator.go#L381

internal/app/admin/http/cache.go
samrabelachew: I moved it to the orchestrator, but I don't believe a lock is necessary. Key and stream deletion all happen under the same cache lock, so the cache and streams are protected from being out of sync. A lock around the entire operation would prevent multiple ClearCache handler calls from running at the same time, but that is not necessary. The DeleteKey() operation is safe against concurrent removals of the same key (we both acquire a lock on the cache before attempting to delete and within first check if the key still exists within the lock).

internal/app/orchestrator/orchestrator.go
```diff
@@ -249,6 +251,17 @@ func (o *orchestrator) GetDownstreamAggregatedKeys() (map[string]bool, error) {
 	return keys, err
 }
 
+func (o *orchestrator) ClearCacheEntries(keys []string) []error {
+	var errors []error
+	for _, key := range keys {
+		_, err := o.cache.DeleteKey(key)
```
jyotimahapatra, line 257: this will remove the entry from the cache, does it also terminate the sidecar streams?

samrabelachew: `DeleteKey()` internally calls `c.cache.Remove(key)` [here](https://github.com/envoyproxy/xds-relay/blob/1c332a20076edbbd5b6250c6c90cea342a5683a9/internal/app/cache/cache.go#L260). The intenral lru Remove() call both removes the element and runs the `onEvicted` func as seen [here](https://github.com/golang/groupcache/blob/master/lru/lru.go#L86-L113). Because when the cache was created, we passed [onEvictCache](https://github.com/envoyproxy/xds-relay/blob/master/internal/app/orchestrator/orchestrator.go#L101) to it, it should clean up any related streams.

internal/app/cache/cache.go
jyotimahapatra: ok thats awesome.

internal/app/admin/http/cache.go
jyotimahapatra: can we avoid using `fmt.Fprintf` to write into response body

internal/app/admin/http/cache.go
```diff
@@ -151,31 +151,25 @@ func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
...
 
-			// Find keys that match prefix of wildcard
-			rootCacheKeyName := strings.TrimSuffix(cacheKey, "*")
-			for potentialMatchKey := range allKeys {
-				if strings.HasPrefix(potentialMatchKey, rootCacheKeyName) {
-					keysToPrint = append(keysToPrint, potentialMatchKey)
-				}
+func clearCacheHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		if req.Method == "POST" {
```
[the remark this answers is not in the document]
jyotimahapatra: may be use a switch case.

internal/app/admin/http/cache.go
```diff
@@ -210,6 +233,16 @@ func printCacheEntries(keys []string, cache cache.ReadOnlyCache, w http.Response
 	}
 }
 
+func clearCacheEntries(keys []string, o *orchestrator.Orchestrator, w http.ResponseWriter) {
+	errors := orchestrator.Orchestrator.ClearCacheEntries(*o, keys)
+	for _, err := range errors {
+		errMessage, _ := stringify.InterfaceToString(&marshallable.Error{
+			Message: err.Error(),
+		})
+		_, _ = w.Write([]byte(errMessage))
```
jyotimahapatra, line 242: i dont think this will create the right response json inside a loop. May be create a string based aggreagated response of errors and put the final message into the Message field

internal/app/admin/http/cache.go
```diff
@@ -210,6 +233,16 @@ func printCacheEntries(keys []string, cache cache.ReadOnlyCache, w http.Response
 	}
 }
 
+func clearCacheEntries(keys []string, o *orchestrator.Orchestrator, w http.ResponseWriter) {
+	errors := orchestrator.Orchestrator.ClearCacheEntries(*o, keys)
```
jyotimahapatra, line 237: can we use `o.ClearCacheEntries(keys)`?

jyotimahapatra (commented):
looks good. few suggestions

internal/app/admin/http/cache.go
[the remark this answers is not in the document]
samrabelachew: Updated it to use `http.MethodPost`, but since it is just a single case to check, I think a switch case isn't really necessary. Instead, I modified it to avoid using an else case. This fxn. was based of the [logHandler](https://github.com/envoyproxy/xds-relay/blob/master/internal/app/admin/http/handler.go#L152-L176), so I can fix the nits of that function in another PR.

internal/app/admin/http/cache.go
samrabelachew: The function provides an orchestrator.Orchestrator pointer, so we would still need to dereference it `(*o).ClearCacheEntries(keys)`. Most of the other handlers run orchestrator functions using `orchestrator.Orchestrator.FUNCTION(*o, ...)`, but I can also change them in a different PR.

internal/app/admin/http/cache.go
```diff
@@ -151,31 +151,25 @@ func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
...
+				clearCacheEntries(keysToClear, o, w)
 			}
-		} else {
-			// Otherwise return the cache entry corresponding to the given key.
-			keysToPrint = []string{cacheKey}
+			return
 		}
-		printCacheEntries(keysToPrint, c, w)
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		_, _ = w.Write([]byte("Only POST is supported\n"))
```
jyotimahapatra, line 172: i think we dont need this error message. `StatusMethodNotAllowed` is self explanatory

internal/app/admin/http/handler_test.go
```diff
@@ -199,6 +199,108 @@ func TestAdminServer_CacheDumpHandler_NotFound(t *testing.T) {
 	assert.Equal(t, "", rr.Body.String())
 }
 
+func TestAdminServer_ClearCacheHandler(t *testing.T) {
```
jyotimahapatra, line 202: what is the difference between tests here and in `internal/app/admin/http/cache_test.go` ?

internal/app/cache/cache.go
```diff
@@ -236,10 +239,29 @@ func (c *cache) DeleteRequest(key string, req transport.Request) error {
 		"node_id", req.GetNodeID(),
 		"request_type", req.GetTypeURL(),
 	).Debug(context.Background(), "request deleted")
-	metrics.CacheAddRequestSubscope(c.scope, key).Counter(metrics.CacheDeleteSuccess).Inc(1)
+	metrics.CacheDeleteRequestSubscope(c.scope, key).Counter(metrics.CacheDeleteSuccess).Inc(1)
 	return nil
 }
 
+func (c *cache) DeleteKey(key string) (Resource, error) {
+	c.cacheMu.Lock()
```
jyotimahapatra, line 247: we've added tests in `internal/app/admin/http/cache_test.go` . Should we add tests for this function as well for completeness?

internal/app/admin/http/handler_test.go
samrabelachew: These tests focus on non-wildcard cases of clearing the cache. I originally left these in `internal/app/admin/http/handler_test.go` since that seemed to be what happened with the cache dump handler tests in a previous PR that split off some cache dump tests into their own file (TestAdminServer_CacheDumpHandler and TestAdminServer_CacheDumpHandler_NotFound remained in handler_test.go while all other cache dump tests were moved to `internal/app/admin/http/cache_test.go`. I went ahead and moved all cache related tests for both admin handlers from the handler file and into the cache file.

internal/app/admin/http/handler.go
```diff
@@ -40,6 +40,12 @@ func getHandlers(bootstrap *bootstrapv1.Bootstrap,
 			readyHandler(weboff),
 			true,
 		},
+		{
+			"/clear_cache",
```
jyotimahapatra, line 44: nit: may be keep it restful by making it `/cache/clear`

jyotimahapatra (approved):

PR #176, opened 2020-10-28
Upstream keepalive

jyotimahapatra:
While debugging a rollout issue in staging, it dawned on us that on some of the xdsrelay pods the grpc streams were not receiving upstream updates. On closer look it looked like this happened on the streams that weren't having a lot of updates. Any stream having regular updates kept healthy.
After reading through https://github.com/grpc/proposal/blob/master/A14-channelz.md , https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md and https://grpc.io/blog/a-short-introduction-to-channelz/ it looked like keepalive time was the suspect.
After trying out upstream keepalive time, the issue was resolved. This PR makes it easy to set the uptream cx keepalive time to a configurable value in the bootstrap. The `Timeout` and `PermitWithoutStream` were set to default of 20s and true. If necessary in the future these can be made configurable too.


internal/app/upstream/client.go
```diff
@@ -360,3 +367,19 @@ func updateConnectivityMetric(ctx context.Context, conn *grpc.ClientConn, scope
 		scope.Gauge(metrics.UpstreamConnected).Update(float64(conn.GetState()))
 	}
 }
+
+func getKeepaliveParams(ctx context.Context, logger log.Logger, c CallOptions) keepalive.ClientParameters {
+	keepaliveClientParams := keepalive.ClientParameters{
+		PermitWithoutStream: true,
```
eapolinario, line 373: do we need to set this?

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -35,10 +34,12 @@ message Server {
     SocketAddress address = 1 [(validate.rules).message.required = true];
 }
 
-// [#next-free-field: 2]
+// [#next-free-field: 3]
 message Upstream {
     // The address for the upstream cluster.
     SocketAddress address = 1 [(validate.rules).message.required = true];
+
+    string keep_alive_time = 2;
```
eapolinario, line 42: can you leave a comment explaining what this is and why its type is `string`?

internal/app/server/server.go
```diff
@@ -175,4 +177,7 @@ func registerEndpoints(ctx context.Context, g *grpc.Server, o orchestrator.Orche
 	clusterservice.RegisterClusterDiscoveryServiceServer(g, gcpv3)
 	endpointservice.RegisterEndpointDiscoveryServiceServer(g, gcpv3)
 	listenerservice.RegisterListenerDiscoveryServiceServer(g, gcpv3)
+
+	// Use https://github.com/grpc/grpc-experiments/tree/master/gdebug to debug grpc channel issues
+	service.RegisterChannelzServiceToServer(g)
```
eapolinario, line 182: nit: do you know if there's any perf implications to having this turned on all the time?

jyotimahapatra: good qs..im not sure of that..since we will be actively monitoring and need tools initially while rolling out, we could make this a bootstrap setting too

internal/app/upstream/client.go
jyotimahapatra: `PermitWithoutStream` allows keepalives to be sent even if there are no streams. This will keep the cx warm and have no cx establishment penalty on the first stream. grpc closes the cx if there are no activew streams and this setting will prohibit that.

api/protos/bootstrap/v1/bootstrap.proto
jyotimahapatra: I`ve noticed a usual way of representing time in human readable format. I can include examples in comment.

internal/app/upstream/client.go
jyotimahapatra: The `ClientParameters` already has the documentation, so i did not repeat them here

api/protos/bootstrap/v1/bootstrap.proto
eapolinario: the fact that the type is string makes me a bit concerned since it has the potential for being misused. How much trouble would it be to make it a `google.protobuf.Duration`?

eapolinario (dismissed):

api/protos/bootstrap/v1/bootstrap.proto
jyotimahapatra: Changing it to `google.protobuf.Duration` is straight forward, but we keepalive will be in minutes or possibly hours and Duration does not give us a good human readable representation apart from strings
From https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Duration 

```
In JSON format, the Duration type is encoded as a string rather than an
 object, where the string ends in the suffix "s" (indicating seconds) and
 is preceded by the number of seconds, with nanoseconds expressed as
 fractional seconds. For example, 3 seconds with 0 nanoseconds should be
 encoded in JSON format as "3s", while 3 seconds and 1 nanosecond should
 be expressed in JSON format as "3.000000001s", and 3 seconds and 1
 microsecond should be expressed in JSON format as "3.000001s".
```

eapolinario (approved):

Issue #179, opened 2020-11-02
Endpoint handler unit tests are flakey


jyotimahapatra:
The endpoint handler tests in the PR are flakey and needs couple of reruns to succeed.
example: https://github.com/envoyproxy/xds-relay/pull/178/checks?check_run_id=1343689679

PR #182, opened 2020-11-08
Resolve flakey admin tests

samrabelachew:
https://github.com/envoyproxy/xds-relay/issues/179

Flakey tests were caused by cache output initial request occasionally missing a version_info field, which was dependent on when the latest response sent through the channel also updates the request struct. Since this isn't crucial to the accuracy of the cache, this PR modifies tests to look at only the key and response of cache output. Tests now unmarshal data into a generic interface and then compare with expected counterparts.


samrabelachew:
@jyotimahapatra @jessicayuen

jyotimahapatra (commented):
Changes lok good. some questions.

internal/app/admin/http/cache_test.go
```diff
@@ -699,6 +501,42 @@ func TestAdminServer_CacheDumpHandler_WildcardSuffix_NotFound(t *testing.T) {
...
+	filecontentsCds, err := ioutil.ReadFile(cdsFile)
+	assert.NoError(t, err)
+	var expectedCdsResponse map[string]interface{}
+	err = json.Unmarshal(filecontentsCds, &expectedCdsResponse)
+	assert.NoError(t, err)
+
+	filecontentsLds, err := ioutil.ReadFile(ldsFile)
+	assert.NoError(t, err)
+	var expectedLdsResponse map[string]interface{}
+	err = json.Unmarshal(filecontentsLds, &expectedLdsResponse)
```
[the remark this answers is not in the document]
samrabelachew: I believe the original struct representing `expectedLdsResponse` is `marshallableResource`, defined [here](https://github.com/envoyproxy/xds-relay/blob/master/internal/app/admin/http/cache.go#L182). `jsonpb.UnmarshalString` expects a proto message, so it wouldn't work.

internal/app/admin/http/testdata/lds_response.json
```diff
@@ -1,39 +0,0 @@
-{
-  "Cache": [
```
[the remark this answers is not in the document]
samrabelachew: Not anymore. The lds_response file is covered by the lds v2 file since we are no longer checking the entire cache.

internal/app/admin/http/cache_test.go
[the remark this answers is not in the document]
jyotimahapatra: sg..can we use the `marshallableResource` struct instead of interface?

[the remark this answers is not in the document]
samrabelachew: I had trouble when doing because the struct contains proto fields `types.Resource`. So, it will fail when attempting to unmarshal those fields using the json function.

internal/app/admin/http/cache_test.go
```diff
@@ -699,6 +505,42 @@ func TestAdminServer_CacheDumpHandler_WildcardSuffix_NotFound(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	actualCacheResponse := actualResponse["Cache"].([]interface{})
+	assert.Equal(t, len(actualCacheResponse), 2)
+	var actualLdsResponse map[string]interface{}
+	var actualCdsResponse map[string]interface{}
+
+	if actualCacheResponse[0].(map[string]interface{})["Key"] == "test_lds" {
+		actualLdsResponse = actualCacheResponse[0].(map[string]interface{})
```
jyotimahapatra, line 531: for my education what is causing the `0` and `1` index difference?

jyotimahapatra: may be we use a for loop so that we arent dependent on exact indexes.

samrabelachew: The order of what gets printed out by the cache depends on the order of keys returned by `GetDownstreamAggregatedKeys()` when first searching through the cache. Since that function returns of map of keys, there is no guaranteed cache output order.

jyotimahapatra: Would this be functionally equivalent ?
```
for c := range actualCacheResponse {
  if c..(map[string]interface{})["Key"] == "test_lds" {
    actualLdsResponse = c.(map[string]interface{})
  } else {
    actualCdsResponse = c.(map[string]interface{})
  }
}
```

samrabelachew: Yes, I can incorporate that change.

internal/app/admin/http/cache_test.go
```diff
@@ -699,6 +505,42 @@ func TestAdminServer_CacheDumpHandler_WildcardSuffix_NotFound(t *testing.T) {
...
+			actualLdsResponse = c.(map[string]interface{})
+		} else {
+			actualCdsResponse = c.(map[string]interface{})
+		}
+	}
+
+	assert.Equal(t, expectedLdsResponse["Key"], actualLdsResponse["Key"])
+	assert.Equal(t, expectedCdsResponse["Key"], actualCdsResponse["Key"])
+	assert.Equal(t, expectedLdsResponse["Resp"], actualLdsResponse["Resp"])
+	assert.Equal(t, expectedCdsResponse["Resp"], actualCdsResponse["Resp"])
```
jyotimahapatra, line 541: one last thing...can we add a check that `Requests`  array has len 1

jyotimahapatra: same with ExpirationTime

samrabelachew: Did it for Requests. Can't with ExpirationTime because it is not an array/slice. I added a NotNil check though

jyotimahapatra (approved):

PR #184, opened 2020-11-10
Add ready endpoint

jyotimahapatra:
While rolling out in staging we realized that there are sometimes cx issues due to which there is cache mismatch between upstream and xdsrelay. Due to the severity of traffic we end up terminating the erroneous pod. We need a way to web off faulty pods to continue investigating the issue.


jessicayuen (approved):
LGTM for the initial PR. We'll need to followup to maintain readiness state.

samrabelachew (approved):

PR #192, opened 2020-11-18
Add endpoint to check stream status

jyotimahapatra:
We are experiencing an edge case where the Response is set to null.
```
➜  xds-relay git:(master) ✗ cs1 -n xdsrelay-staging exec -it xdsrelay-staging-6779b775f6-nh6nx -c xdsrelay-service-gojson -- curl "0:6070/cache/v3-metricsdiscoveryengine-staging-iad_cds"
{
  "Cache": [
    {
      "Key": "v3-metricsdiscoveryengine-staging-iad_cds",
      "Resp": null,
      "Requests": null,
      "ExpirationTime": "2020-11-18T21:37:40.045382228Z"
    }
  ]
}

```
There are currently 2 hypothesis
1. There is a race condition bug in orchstrator, due to which the stream is not created with the upstream control plane
2. The stream is actually present and is broken or overriding nil responses.

This endpoint will help us understand the current status of the stream and the last known version on it.


internal/app/upstream/client.go
```diff
@@ -51,6 +51,11 @@ type Client interface {
 	// All goroutines that depend on the ctx object should consider ctx.Done to be related to shutdown.
 	// All such scenarios need to exit cleanly and are not considered an erroneous situation.
 	OpenStream(transport.Request, string) (<-chan transport.Response, func())
+
+	// GetStream is used as a debugging tool to get the latest state of the stream for an aggregated key
+	// If the aggregated key is absent, then either the stream was not created or was already cancelled.
+	// If present, returns the current version of the response from upstream
+	GetStream(string) (string, error)
```
jessicayuen, line 58: Is the version enough or would this be more useful for debugging if we returned the entire response? `GetStreamLatestVersion` might be more accurate.

jyotimahapatra: the version is enough IMO to know the latest on the stream. Will change the name.

jessicayuen (approved):

PR #204, opened 2020-11-25
update gcp + add benchmark test for createWatch + remove usage of maps key'ed on unique DiscoveryRequests

jyotimahapatra:


internal/app/orchestrator/downstream.go
```diff
@@ -14,82 +14,73 @@ package orchestrator
 import (
 	"sync"
 
-	"github.com/envoyproxy/xds-relay/internal/app/mapper"
 	"github.com/envoyproxy/xds-relay/internal/app/transport"
 )
 
 // downstreamResponseMap is a map of downstream xDS client requests to response
 // channels.
```
jessicayuen, line 21: The function comments in this file needs an update.

internal/app/orchestrator/downstream.go
```diff
@@ -14,82 +14,73 @@ package orchestrator
...
 type downstreamResponseMap struct {
 	mu      sync.RWMutex
-	watches map[transport.Request]transport.Watch
+	watches map[string][]transport.Watch
```
jessicayuen, line 24: For performance reasons, I think it's ok we sign off on this approach. My only concern is we lose some debug information that comes from the having the request available.  Ex: connected nodes to a particular key. But perhaps this is available in the watch?

internal/app/orchestrator/downstream.go
```diff
@@ -14,82 +14,73 @@ package orchestrator
...
 // createWatch initializes a new channel for a request if it doesn't already
 // exist.
-func (d *downstreamResponseMap) createWatch(req transport.Request) transport.Watch {
+func (d *downstreamResponseMap) addWatch(aggregatedKey string, w transport.Watch) {
 	d.mu.Lock()
 	defer d.mu.Unlock()
-	if _, ok := d.watches[req]; !ok {
-		d.watches[req] = req.CreateWatch()
+	if _, ok := d.watches[aggregatedKey]; !ok {
+		d.watches[aggregatedKey] = make([]transport.Watch, 0, 10)
```
jessicayuen, line 39: What's the reasoning behind the 10 buffer choice here?

internal/app/orchestrator/downstream.go
```diff
@@ -14,82 +14,73 @@ package orchestrator
...
 
 // get retrieves the channel where responses are set for the specified request.
-func (d *downstreamResponseMap) get(req transport.Request) (transport.Watch, bool) {
+func (d *downstreamResponseMap) getSnapshot(aggregatedKey string) ([]transport.Watch, bool) {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
-	channel, ok := d.watches[req]
-	return channel, ok
+	watches, ok := d.watches[aggregatedKey]
+	if ok {
```
jessicayuen, line 49: Should this be `if !ok {`? If yes, may want to add a test that captures this use case.

internal/app/orchestrator/downstream.go
```diff
@@ -14,82 +14,73 @@ package orchestrator
...
-	defer d.mu.Unlock()
-	if watch, ok := d.watches[req]; ok {
-		// wait for all writes to the responseChannel to complete before closing.
-		watch.Close()
-		delete(d.watches, req)
-		return watch
-	}
-	return nil
+func (d *downstreamResponseMap) delete(w transport.Watch) {
+	w.Close()
```
jessicayuen, line 66: Do we need to remove it from the slice as well?

internal/app/orchestrator/orchestrator.go
```diff
@@ -143,7 +139,9 @@ func (o *orchestrator) CreateWatch(req transport.Request) (transport.Watch, func
 
 		// TODO (https://github.com/envoyproxy/xds-relay/issues/56)
 		// Support unnaggregated keys.
-		return o.downstreamResponseMap.delete(req), nil
+		// nil is a kill pill which will cause go-control-plane to error out the stream
+		_ = o.send(nil, watch, aggregatedKey)
```
jessicayuen, line 143: Would just be explicit here and do `_ = watch.Send(nil)`, since if `Send` errors it implies a fanout problem.

internal/app/orchestrator/orchestrator.go
```diff
@@ -191,24 +179,24 @@ func (o *orchestrator) CreateWatch(req transport.Request) (transport.Watch, func
 		o.logger.With("error", err).With("aggregated_key", aggregatedKey).Warn(ctx, "failed to fetch aggregated key")
 	}
 
+	cachedResponse := false
```
jessicayuen, line 182: is this the same as the `cached.Resp != nil` check that we use in the next line? Can we just reuse that rather than introduce a new var?

internal/pkg/log/log_mock.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 // MockLogger mocks a very basic debug logger.
-var MockLogger = New("debug", os.Stderr)
+var MockLogger = New("error", os.Stderr)
```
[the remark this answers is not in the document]
jyotimahapatra: there was lot of logs during test. I felt we can make it more verbose when things fail. Its hard to see the failures in CI due to so much of logs

internal/app/orchestrator/downstream.go
jyotimahapatra: Yes. The tradeoff here is to keep the core path simple...we should add the deuggability feature as a map as long as we are not looping over the map entries.

internal/app/orchestrator/downstream.go
jyotimahapatra: it can be 1. I was experimenting with different initial size allocations to see their impact on benchmark

internal/app/orchestrator/downstream.go
jyotimahapatra: this is expected..if there was an entry, the snapshot is replaced. If its not found, then no action needed. Added tests for it

internal/app/orchestrator/downstream.go
jyotimahapatra: we dont do it since removal from slice is expensive. The slice is completely rolled over during fanout.

internal/app/orchestrator/orchestrator.go
jyotimahapatra: `o.send` is just a wrapper over `watch.Send`. We do good amount of logging and metrics on channel full and I didn't want to repeat it all over in orchestrator

internal/app/orchestrator/orchestrator.go
jyotimahapatra: thats right..changed it to use a single variable.

PR #205, opened 2020-12-01
update gcp

jyotimahapatra:
Updating go-control-plane until envoyproxy/go-control-plane@8f9846e


jyotimahapatra:
@jessicayuen This change looks promising. Its on staging and working well now. PTAL!

PR #206, opened 2020-12-02
Dequeue stale responses

jessicayuen:
When sending the response to clients, drop the older response currently
in the transport queue, so that we always send the latest response.

Also,
* Add stats scope to watch / transport layer
* Return nil in unaggregated key scenario

### DIFF
--- a/internal/app/transport/requestv2.go
+++ b/internal/app/transport/requestv2.go
@@ -3,6 +3,7 @@ package transport
 import (
 	discoveryv2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
 	structpb "github.com/golang/protobuf/ptypes/struct"
+	"github.com/uber-go/tally"
 	status "google.golang.org/genproto/googleapis/rpc/status"
 )
 
@@ -79,6 +80,6 @@ func (r *RequestV2) GetResponseNonce() string {
 }
 
 // CreateWatch creates a versioned Watch
-func (r *RequestV2) CreateWatch() Watch {
-	return newWatchV2()
+func (r *RequestV2) CreateWatch(scope tally.Scope) Watch {
+	return newWatchV2(scope)
 }
--- a/internal/app/transport/watch_test.go
+++ b/internal/app/transport/watch_test.go
@@ -12,6 +12,8 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+
+	"github.com/envoyproxy/xds-relay/internal/pkg/stats"
 )
 
 type version int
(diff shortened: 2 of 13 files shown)


jyotimahapatra:
https://github.com/envoyproxy/xds-relay/pull/205 might have conflicts with changes in this PR.

jessicayuen:
@jyotimahapatra Which one do we want to merge first?

samrabelachew (dismissed):

internal/app/transport/watchv2.go
```diff
@@ -43,10 +49,22 @@ func (w *watchV2) Send(s Response) error {
 	if w.closed {
 		return nil
 	}
+
+	// Drop the older response currently in the transport queue, so that we
```
jyotimahapatra, line 53: I wanted to brainstorm an alternative approach based on the version of gcp we use https://github.com/envoyproxy/go-control-plane/blob/53deebd8777dd83921cb83a1975e7ec13fb8bb04/pkg/server/sotw/v3/server.go#L200

If instead of deque/enqueue operation, if we closed the channel , then the stream will be terminated and the envoy sidecar will create a fresh stream. I think this will help keep the implementation simple. wdyt?

jyotimahapatra: Another aspect to think about is : The channel drop happens when the transport is blocked, which can only happen when there is stress in the system(cpu/network). One way to apply backpressure in such situation would be to close the stream so that the sidecar will recreate the stream(which can load balance and hit another pod).

jyotimahapatra: The PR approach is fine too and we can go ahead with it, just wanted to make sure we add extra code with the documented reasons.

jessicayuen: I think it depends if we want to cause thrashing at the transport layer (clients -> control plane) or have the control plane alleviate that burden. It depends on the setup of the clients IMO. Imagine a scenario if the clients all reside in a single environment (ex: a Kubernetes cluster) with xds-relay in a different environment. If the client environment is degraded and we started closing watches, we'll see a churn of new connections being established to the control plane. With the current approach, we essentially hold the connection until the clients are ready to receive updates again.

jyotimahapatra: 'we'll see a churn of new connections being established to the control plane` there will be stream churn and not connection. The current approach is fine..i was trying to see if we can achieve same effect with less code

internal/app/transport/requestv2.go
```diff
@@ -79,6 +80,6 @@ func (r *RequestV2) GetResponseNonce() string {
 }
 
 // CreateWatch creates a versioned Watch
-func (r *RequestV2) CreateWatch() Watch {
-	return newWatchV2()
+func (r *RequestV2) CreateWatch(scope tally.Scope) Watch {
+	return newWatchV2(scope)
```
jyotimahapatra, line 84: in a future PR, we might need to separate watch from request

internal/app/transport/watch_test.go
```diff
@@ -23,8 +25,13 @@ const (
 
 var discoveryResponsev2 = &discoveryv2.DiscoveryResponse{}
 var discoveryResponsev3 = &discoveryv3.DiscoveryResponse{}
+var discoveryResponse2v2 = &discoveryv2.DiscoveryResponse{TypeUrl: "type.googleapis.com/envoy.api.v2.Listener"}
```
jyotimahapatra, line 28: nit: could use resourcev2.ListenerType`

jyotimahapatra (dismissed):
lgtm

jyotimahapatra (approved):

Issue #144, opened 2020-09-14
Verbose and succinct cache output


jyotimahapatra:
While debugging caches i realized the cache output is too big because it includes the DiscoveryRequest.
This was hard to work with. We should make a way to show verbose/non verbose logs on cache endpoint.

samrabelachew:
@jyotimahapatra @jessicayuen, what do you think of this?
Verbose/succinct setup:
Update cache handler to also accept a `-v` flag.

When `-v` flag is used, existing cache flow is run to output the entire cache.
When `-v` flag is excluded, cache output becomes a summary as follows:
```
Cache: [
	{
	"aggkey1": {
		"resp_typeURL”: string,
		"resp_version”: int,
		"num_requests: int,
		"exp_time”: string
		}
	}
]
```

jessicayuen:
How about using query params for the verbose cache option, ex: `/cache/<key>?verbose=true`

jessicayuen:
The succinct cache output looks good, except I would return the full response rather than just the typeURL/version. What are your thoughts? cc @jyotimahapatra

PR #208, opened 2020-12-02
Verbose cache

samrabelachew:
Per https://github.com/envoyproxy/xds-relay/issues/144, made cache handler have a verbose/succinct output. Default is to run succinctly. If URL contains parameter `verbose=true`, it will run the original verbose method, including all request metadata. Otherwise, it will return the number of requests in its place.


samrabelachew:
@jyotimahapatra @jessicayuen

internal/app/admin/http/cache.go
```diff
@@ -265,13 +267,18 @@ func resourceToPayload(key string, resource cache.Resource) []marshallableResour
...
 		Resp:           marshalDiscoveryResponse(resource.Resp),
-		Requests:       requests,
 		ExpirationTime: resource.ExpirationTime,
-	})
+	}
+
+	if isVerbose {
+		marshallableResource.Requests = requests
+	} else {
+		marshallableResource.NumRequests = len(requests)
```
jyotimahapatra, line 279: `NumRequests` can always be present. It wont affect the payload size

jessicayuen:
@samrabelachew can you merge in the conflicts when you get the chance and I'll relook at this?

internal/app/admin/http/handler.go
```diff
@@ -165,6 +167,15 @@ func getParam(path string, prefix string) string {
 	return param
 }
 
+func getQueryValue(values url.Values) bool {
```
jessicayuen, line 170: Can we make this more generic for returning query values? ex:
```
func getBoolQueryValue(values url.Values, key string) (bool, err) {
...
}
```

jessicayuen (commented):
Thanks, lgtm! Just one comment.

internal/app/admin/http/handler.go
```diff
@@ -165,6 +167,15 @@ func getParam(path string, prefix string) string {
 	return param
 }
 
+func getBoolQueryValue(values url.Values, key string) (bool, error) {
+	verbose := values.Get(key)
+	isVerbose, err := strconv.ParseBool(verbose)
+	if err == nil {
+		return isVerbose, nil
+	}
+	return false, err
```
jessicayuen, line 176: Can we remove the `verbose` from here too? This can be shortened to the following since Go defaults for bool will be false:
```
	val := values.Get(key)
	return strconv.ParseBool(val)
```

jessicayuen (approved):

PR #209, opened 2020-12-03
Version Handler

samrabelachew:
Added version handler to print version of response in cache given a key.


samrabelachew:
@jessicayuen @jyotimahapatra

internal/app/admin/http/cache.go
```diff
@@ -106,6 +106,38 @@ func edsDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 	}
 }
 
+func versionHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		cdsKey := filepath.Base(req.URL.Path)
```
jyotimahapatra, line 111: nit: drop `cds` in variable names

internal/app/admin/http/cache.go
```diff
@@ -106,6 +106,38 @@ func edsDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
...
+			w.WriteHeader(http.StatusNotFound)
+			s, _ := stringify.InterfaceToString(&marshallable.Error{
+				Message: err.Error(),
+			})
+			_, _ = w.Write([]byte(s))
+			return
+		}
+
+		version := resp.Resp.GetPayloadVersion()
+		_, e := w.Write([]byte(version))
```
jyotimahapatra, line 132: do we want to wrap the response in an object? We could choose to add more fields later without breaking the clients.

jyotimahapatra (approved):

PR #213, opened 2020-12-09
Clear Cache Endpoint Params

samrabelachew:
Clear cache handler should have a path consisting of 4 entries (`address/cache/clear/x`). This change has it parse its own parameter instead of relying on `getParam()`

### DIFF
--- a/internal/app/admin/http/cache.go
+++ b/internal/app/admin/http/cache.go
@@ -189,7 +189,7 @@ func keyDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 
 func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
-		cacheKey := getParam(req.URL.Path, "/cache")
+		cacheKey := getParam(req.URL.Path, cacheUrl)
 		c := orchestrator.Orchestrator.GetReadOnlyCache(*o)
 		keysToPrint, err := getRelevantKeys(o, cacheKey, w)
 		if err == nil {
@@ -201,7 +201,7 @@ func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 func clearCacheHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
 		if req.Method == http.MethodPost {
-			cacheKey := getParam(req.URL.Path, "/cache/clear")
+			cacheKey := getParam(req.URL.Path, clearUrl)
 			keysToClear, err := getRelevantKeys(o, cacheKey, w)
 			if err == nil {
 				clearCacheEntries(keysToClear, o, w)
--- a/internal/app/admin/http/cache.go
+++ b/internal/app/admin/http/cache.go
@@ -201,7 +201,11 @@ func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 func clearCacheHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
 		if req.Method == http.MethodPost {
-			cacheKey := getParam(req.URL.Path)
+			var cacheKey string
+			splitPath := strings.SplitN(req.URL.Path, "/", 4)
+			if len(splitPath) == 4 {
+				cacheKey = splitPath[3]
+			}
 			keysToClear, err := getRelevantKeys(o, cacheKey, w)
 			if err == nil {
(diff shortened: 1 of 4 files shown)


internal/app/admin/http/cache.go
```diff
@@ -201,7 +201,11 @@ func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 func clearCacheHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
 		if req.Method == http.MethodPost {
-			cacheKey := getParam(req.URL.Path)
+			var cacheKey string
```
jyotimahapatra, line 204: can we instead edit `getParam` https://github.com/envoyproxy/xds-relay/blob/8a16fd9860e4b65f51e8194e04ce22f16ead297b/internal/app/admin/http/handler.go#L157-L160 to return the last segment irrespective of length?

samrabelachew: Depends, I avoided doing this initially since we allowed defining `/cache` as equivalent to `/cache/` (i.e. print total cache). The log level handler also has meaning behind omitting the trailing slash. If we return final segment irrespective of length, this would cause a bug. Alternatively, we can require the final trailing slash for all handlers using `getParam` and this would work.

jyotimahapatra: how about using this https://github.com/envoyproxy/xds-relay/blob/eda7072ba1dee995cafa23e5f63d519cf63787c3/internal/app/admin/http/cache.go#L34

samrabelachew: I instead modified getParam() to take in a prefix to remove from the Path and used filepath.Split() to collect the remaining argument or empty string. I chose not to use Base() since it would return a `.` should the remaining path be empty.

internal/app/admin/http/cache.go
```diff
@@ -189,7 +189,7 @@ func keyDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 
 func cacheDumpHandler(o *orchestrator.Orchestrator) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
-		cacheKey := getParam(req.URL.Path)
+		cacheKey := getParam(req.URL.Path, "/cache")
```
jyotimahapatra, line 192: can we make the url prefixes constants?

internal/app/admin/http/cache.go
jyotimahapatra: this approach lgtm. Can you make the urls constants so that there is less chance of bugs due to future changes.

internal/app/admin/http/handler.go
```diff
@@ -23,6 +24,10 @@ type Handler struct {
 	redirect    bool
 }
 
+const logUrl = "/log_level"
```
jyotimahapatra, line 27: nit: may be use a block to define constants https://github.com/envoyproxy/xds-relay/blob/master/internal/app/metrics/metrics.go#L11-L15

internal/app/admin/http/cache_test.go
```diff
@@ -485,106 +485,108 @@ func testAdminServerCacheDumpHelper(t *testing.T, urls []string) {
...
+			}
+			upstreamResponseChannelCDS <- &resp
+			gotResponse = <-cdsRespChannel.GetChannel().V2
+			gotDiscoveryResponse, err = gotResponse.GetDiscoveryResponse()
+			assert.NoError(t, err)
+			assert.Equal(t, &resp, gotDiscoveryResponse)
 
-		req, err := http.NewRequest("GET", url, nil)
-		assert.NoError(t, err)
+			req, err := http.NewRequest("GET", url, nil)
```
jyotimahapatra, line 577: nit: fix  indentation.

samrabelachew: Indentation comes from wrapping test in a `t.Run()` and running `golangci-lint`

jyotimahapatra: but this indentation is not intended, no? in that case can you manually remove the unnecessary indentation?

samrabelachew: I think it's intended to keep it in line with the new indentation https://github.com/envoyproxy/xds-relay/blob/21d855083faecfe9f20be54f81b7d761fbb5055f/internal/app/admin/http/cache_test.go#L567-L589

(I can still remove it, but I think it will fail unit test lint checks if I do)

jyotimahapatra (approved):

internal/app/admin/http/cache_test.go
jyotimahapatra: i think i have an issue in my github diff view. The indentation is fine

PR #214, opened 2020-12-17
Implement per stream timeout

pcalley:
Allows `timeout` parameter to be optionally added under `origin_server` in the bootstrap config in order to add a per stream timeout for each connection to the upstream server. 

Timeout is a duration in milliseconds and defaults to no timeout. 

Jitter of timeout values to come in subsequent PR.


internal/app/upstream/client_test.go
```diff
@@ -358,6 +362,62 @@ func TestOpenStreamShouldRetryIfSendFails(t *testing.T) {
...
+		var val int64 = 0
+		if scope.Snapshot().Counters() != nil &&
+			scope.Snapshot().Counters()["mock.lds.stream_retry+key=aggregated_key"] != nil {
+			val = scope.Snapshot().Counters()["mock.lds.stream_retry+key=aggregated_key"].Value()
+		}
+		if val > 0 {
+			break
+		}
+	}
+	done()
```
samrabelachew, line 389: One question for my understanding: why do we call `defer done()` if we run `done()` here?

samrabelachew (changes requested):

api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -47,6 +47,10 @@ message Upstream {
     // From https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Duration
     // Duration only lets us represent time in 's'
     string keep_alive_time = 2;
+
+    // Timeout in milliseconds for upstream connection stream.
+    // If unset defaults to no timeout.
+    int64 timeout = 3;
```
jessicayuen, line 53: Can you explain how this differentiates from the the `keep_alive_time` above?

pcalley: The `keep_alive_time` tells the stream to send a ping (defaults every 5m) and would close any broken streams. The `timeout` cancels the context that each stream uses on a fixed interval forcing the connection to close and reopen. My assumption is that when the stream is cancelled it forces parts of the cache to be refreshed.

internal/app/upstream/client_test.go
pcalley: Good catch I think the second done is not necessary.

api/protos/bootstrap/v1/bootstrap.proto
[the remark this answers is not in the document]
pcalley: Fixed.

internal/app/upstream/client_test.go
```diff
@@ -358,6 +362,60 @@ func TestOpenStreamShouldRetryIfSendFails(t *testing.T) {
...
+	for {
+		var val int64 = 0
+		if scope.Snapshot().Counters() != nil &&
+			scope.Snapshot().Counters()["mock.lds.stream_retry+key=aggregated_key"] != nil {
+			val = scope.Snapshot().Counters()["mock.lds.stream_retry+key=aggregated_key"].Value()
+		}
+		if val > 0 {
+			break
+		}
+	}
```
jessicayuen, line 388: If the test condition fails, this loop will cause the test to hang indefinitely is that correct? Can we make this test automatically fail if a set duration has passed?

jessicayuen (commented):
LGTM to me, just the one outstanding comment.

internal/app/upstream/client_test.go
pcalley: Yes this is correct. The latest commit should check for this condition and fail the test after 5s if the success condition isn't met.

jessicayuen (approved):

samrabelachew (approved):

PR #215, opened 2020-12-30
Fix cache key in README

nak3:
This patch fixes to:
- add `--base-id {0,1}` option to run multiple envoys on the same host.
- replaces `cache/cluster1_cds` with `cache/staging_cds`. It was updated by cd675d4.


jessicayuen (approved):
Thanks!

PR #218, opened 2021-01-04
Fix log format

nak3:
This patch make a tiny change which fixes a log format.

BFORE:
```
{"level":"info","ts":1609729155.4176533,"msg":"received interrupt signal:%!(EXTRA string=interrupt)","json":{}}
```

AFTER:
```
{"level":"info","ts":1609806458.4018972,"msg":"received interrupt signal","json":{"signal":"interrupt"}}
```

Signed-off-by: Kenjiro Nakayama <nakayamakenjiro@gmail.com>

### DIFF
--- a/internal/app/server/server.go
+++ b/internal/app/server/server.go
@@ -164,7 +164,7 @@ func RunWithContext(ctx context.Context, cancel context.CancelFunc, bootstrapCon
 				stopMutex.Unlock()
 			}
 		case sig := <-sigs:
-			logger.Info(ctx, "received interrupt signal:", sig.String())
+			logger.With("signal", sig.String()).Info(ctx, "received interrupt signal")
 			logger.Info(ctx, "initiating admin server shutdown")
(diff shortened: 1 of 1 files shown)


internal/app/server/server.go
```diff
@@ -164,7 +164,7 @@ func RunWithContext(ctx context.Context, cancel context.CancelFunc, bootstrapCon
 				stopMutex.Unlock()
 			}
 		case sig := <-sigs:
-			logger.Info(ctx, "received interrupt signal:", sig.String())
+			logger.Info(ctx, "received interrupt signal: %s", sig.String())
```
jessicayuen, line 167: Thanks for the fix. Could you use `logger.With("signal", sig.String()).Info(ctx, "received interrupt signal")`?

It allows logging engines (ex: Kibana) to filter on the exact message and context cues.

nak3: Sure, updated. Thank you!

jessicayuen (approved):
Thanks!

PR #222, opened 2021-01-14
Add jitter for per stream timeouts

pcalley:
Add Jitter in seconds for upstream connection. Used with timeout to reset connections without overloading the upstream server with reconnect requests. Jitter is the upper tolerance for random variation in the timeout. 
e.g. timeout=15, jitter=5 -> connectionTimeout is rand Int 15 and 20. Defaults to no jitter.


api/protos/bootstrap/v1/bootstrap.proto
```diff
@@ -51,6 +51,13 @@ message Upstream {
     // Timeout in seconds for upstream connection stream.
     // If unset defaults to no timeout.
     int64 timeout = 3;
+
+    // Jitter in seconds for upstream connection. Used with timeout to reset connections without overloading
+    // the upstream server. Jitter is the lower tolerance for random variation in the timeout. e.g. timeout=15,
+    // jitter=5 -> connectionTimeout is rand Int 10 and 15.
```
jessicayuen, line 57: It might make more sense if jitter is the upper tolerance since timeouts should generally be respected. If the connection isn't as long-lived as users expect, we may run into unexpected behavior from the client.

internal/app/upstream/client.go
```diff
@@ -175,7 +179,12 @@ func (m *client) handleStreamsWithRetry(
...
 		if m.timeout != 0 {
-			childCtx, cancel = context.WithTimeout(ctx, time.Duration(m.timeout)*time.Second)
+			timeout := m.timeout
+			if m.jitter > 0 && m.jitter < timeout {
+				timeout = m.timeout - rand.Int63n(m.jitter)
+			}
+			m.logger.With("aggregated_key", aggregatedKey).Debug(ctx, "Connecting to upstream with timeout: %ds", timeout)
+			childCtx, cancel = context.WithTimeout(ctx, time.Duration(timeout)*time.Second)
 		} else {
 			childCtx, cancel = context.WithCancel(ctx)
```
jessicayuen, line 189: Could you also add a log line here that says connecting to upstream? Just to ensure we have parity since you add the message with the timeout above.

jessicayuen (changes requested):
LGTM! Just a couple small nits:

samrabelachew (commented):
Agree w/Jess's suggestions. Otherwise lgtm!

internal/app/upstream/client.go
pcalley: done

jessicayuen (approved):

PR #223, opened 2021-01-14
upstream client: Handle recv() blocked scenarios

jessicayuen:
There are two error scenarios currently that leads to cache drifts
between the management server and xds-relay.

1. There is a bug where the stream is stuck indefinitely if the send chan
is closed unexpectedly. The recv routine stays open and the waitgroup
never completes, blocking retries.

2. If the response receiver channel for `recv()` is blocked, the upstream
client is blocked from sending ACKs back to the management server.


samrabelachew (approved):
lgtm

